Issue #1, opened 2018-01-22
Repository should have a license file


axic:

ehildenb:
@axic please close if you're happy with the NCSA license

PR #2, opened 2018-01-22
LICENSE: adding NCSA/UIUC license

ehildenb:
Addressing: https://github.com/kframework/wasm-semantics/issues/1

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,60 @@
+==============================================================================
+The WASM Semantics in K Release License
+==============================================================================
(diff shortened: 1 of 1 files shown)


daejunpark:
@ehildenb should be essentially the same with evm-semantics, right?

daejunpark:
@grosu you may want to take a look at this as well.

ehildenb:
Yes, I copied the license from the EVM semantics, and replaced `EVM` with `WASM`

PR #5, opened 2018-02-23
New tangle

ehildenb:


ehildenb:
@bmmoore see for example on how to use the `pandoc-tangle` repo as a submodule in build process

PR #8, opened 2018-03-07
Stack ops

ehildenb:


ehildenb:
@chfast please review. Ignore the first two organizational commits.

wasm-syntax.md
```diff
@@ -173,8 +173,8 @@ When a binary operator is the next instruction, `#eval<T>BinOp` will be called o
  // ---------------------------------------------------------------
 ```
 
-Arithmetic
-----------
+Numeric Operators
```
chfast: For sanity, shouldn't it also be `### Numeric Operators`?

wasm-syntax.md
```diff
@@ -34,14 +34,18 @@ WASM values are either integers or floating-point numbers, of 32- or 64-bit widt
 First, the type of values is built, then values annotated with their types are provided.
 
 ```k
+    syntax Data ::= Int | Float
```
chfast: What is `Data`. A new type? Would it make sense to name it `Number`?

tests/stack/polymorphic.wast
```diff
@@ -0,0 +1,43 @@
...
+drop
+#assertStack .Stack "drop f64"
+
+// select
+
+i32.const -1
+i32.const 1
+i32.const 1
+select
+#assertTopStack < i32 > 1 "select i32 true"
```
chfast: Where are the `true`/`false` literals coming from?

ehildenb: `"true"` here is just part of the string. The `#assertTopStack` is an extra thing I put in for the testing harness, the string (which is last/second argument) is just for naming the test.

In this case, it should select the "true" branch of the `select`, which means keep the top element of the stack instead of the second element.

wasm-syntax.md
ehildenb: `Data` is a new type added as a supertype of `Int` and `Float`. I think `Number` does sound better, will change.

wasm-syntax.md
ehildenb: I'm trying to mimic the structure of the WASM spec. Numeric Operators starts on page 63. I think it's best to wait until a bit more of the semantics is finished, then I'll do a re-organizational PR to make it mimic the WASM spec much more.

ehildenb:
@chfast I changed `Data => Number`, but I think will leave the documentation related changes until a later PR. Is it ok?

PR #9, opened 2018-03-07
Labels control flow

ehildenb:
This pull request introduces labels in the semantics, as well as the structured control-flow instructions which use them. It does not implement the inter-function instructions (like `call*`, `invoke`, or `return`).

### DIFF
--- a/tests/stack/control-flow.wast
+++ b/tests/stack/control-flow.wast
@@ -121,3 +121,11 @@ if [ i32 ] i32.const 1 else i32.const -1 end
 //     i32.add
 // end
 // #assertTopStack < i32 > 10 "loop"
+
+// Stack Underflow
+// TODO: We need to give semantics to stack underflow (though it could not happen with a validated program).
+// We need `trap` semantics first.
+// i32.const 0
+// block [ i32 i32 ]
+//     i32.const 7
+// end
--- a/tests/stack/control-flow.wast
+++ b/tests/stack/control-flow.wast
@@ -29,3 +29,11 @@ block [ ]
     br 0
 end
 #assertStack < i32 > 2 : < i32 > 1 : .Stack "block 4"
+
+i32.const 1
+if [ i32 ] i32.const 1 else i32.const -1 end
+#assertTopStack < i32 > 1 "if true"
+
+i32.const 0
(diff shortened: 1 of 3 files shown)


ehildenb:
@axic please review

wasm-syntax.md
```diff
@@ -28,25 +28,27 @@ $2^32$ and $2^64$ are used often enough to warrant providing macros for them.
...
     syntax  ValType ::= IValType | FValType
  // ---------------------------------------
 
-    syntax IVal ::= "<" IValType ">" Int
-    syntax FVal ::= "<" FValType ">" Float
-    syntax  Val ::= "<"  ValType ">" Number
-                  | IVal | FVal
- // ---------------------------
+    syntax ValTypes ::= List{ValType, ""}
+    syntax ValType  ::= "[" ValTypes "]"
```
axic: It is a bit confusing that this allows nested types: `[ [ i32 [ i32 ] [ i32 i32 i32 ] ] i32 ]`

Is there a case where this is needed? If not, perhaps restricting this to a single vector is nicer.

axic:
I think with discussion with @ehildenb I've managed to understand these changes and they look good.

tests/stack/control-flow.wast
```diff
@@ -0,0 +1,123 @@
...
+    drop
+end
+#assertStack < i32 > 2 : < i32 > 1 : .Stack "block 2"
+
+block [ i32 i32 ]
+    i32.const 1
+    i32.const 2
+    i32.const 3
+end
+#assertStack < i32 > 3 : < i32 > 2 : .Stack "block 3 (invalid)"
```
axic: Should also add a stack underflow case:
```
block [ i32 i32 ]
   i32.const 1
end
```

wasm-syntax.md
ehildenb: I have adjusted the syntax to only allow one level of nesting, which is tighter to the WASM spec.

tests/stack/control-flow.wast
ehildenb: I've added a TODO, and an issue https://github.com/kframework/wasm-semantics/issues/11

ehildenb:
@axic changes look ok? I made `VecType` and a TODO about stack underflow tests.

PR #10, opened 2018-03-12
Frames

ehildenb:


ehildenb:
@pirapira this is ready for review once @axic reviews the previous PR (so that you don't have to review as much).

ehildenb:
@pirapira ready for review

wasm.md
```diff
@@ -135,6 +296,36 @@ These functions make assertions about the state of the `<stack>` cell.
...
 ```
 
+The operator `#assertLocal`/`#assertGlobal` operators perform a check for a local/global variable's value.
+
+```k
+    syntax Instr ::= "#assertLocal"  Int Val String
+                   | "#assertGlobal" Int Val String
+ // -----------------------------------------------
+    rule <k> #assertLocal INDEX VALUE _ => . ... </k>
+         <locals> ... (INDEX |-> VALUE => .Map) ... </locals>
```
pirapira: I didn't expect `#assertLocal` to have a side effect of clearing a pair from locals.

ehildenb: Hmmmm, this is a good point, especially if we want to expose the assertion language to users of KWASM for lightweight runtime verification of programs they write.

The reason I have the various `#assert...` statements clear the part of the state they are asserting about is so that at the end of running a test, the configuration is completely empty and can just be compared against `tests/stack/success.out` to check that the entire test-file "passed".

I could instead add a `#clear` operator, which clears the entirety of the state in one step. A similar thing has been added to the KEVM semantics.

pirapira: In unit tests, it feels clean when everything is assert-checked and removed from the configuration.

Some people are more used to assertions that can be turned on or off.

Issue #21, opened 2018-12-04
Makefile archiving problem


MrChico:
On my MacOS machine, `make build` results in the following error even though the individual steps of the building process succeeds. It's probably due to some arcane Makefile magic I can't figure out:
```
➜  wasm-semantics git:(function-semantics) ✗ make build
==  tangle: ~/wasm-semantics/.build/defn/wasm/wasm.k
mkdir -p ~/wasm-semantics/.build/defn/wasm/
pandoc --from markdown --to .build/pandoc-tangle/tangle.lua --metadata=code:.k wasm.md > ~/wasm-semantics/.build/defn/wasm/wasm.k
==  tangle: ~/wasm-semantics/.build/defn/wasm/data.k
mkdir -p ~/wasm-semantics/.build/defn/wasm/
pandoc --from markdown --to .build/pandoc-tangle/tangle.lua --metadata=code:.k data.md > ~/wasm-semantics/.build/defn/wasm/data.k
== kompile: ~/wasm-semantics/.build/defn/wasm/wasm-kompiled/interpreter
eval $(opam config env) \
	~/wasm-semantics/.build/k/k-distribution/target/release/k/bin/kompile --debug --gen-ml-only -O3 --non-strict \
					 --main-module WASM --syntax-module WASM ~/wasm-semantics/.build/defn/wasm/wasm.k --directory ~/wasm-semantics/.build/defn/wasm \
		&& ocamlfind opt -c ~/wasm-semantics/.build/defn/wasm/wasm-kompiled/constants.ml -package gmp -package zarith \
		&& ocamlfind opt -c -I ~/wasm-semantics/.build/defn/wasm/wasm-kompiled \
		&& ocamlfind opt -a -o ~/wasm-semantics/.build/defn/wasm/semantics.cmxa \
		&& ocamlfind remove wasm-semantics-plugin \
		&& ocamlfind install wasm-semantics-plugin META ~/wasm-semantics/.build/defn/wasm/semantics.cmxa ~/wasm-semantics/.build/defn/wasm/semantics.a \
		&& ~/wasm-semantics/.build/k/k-distribution/target/release/k/bin/kompile --debug --packages wasm-semantics-plugin -O3 --non-strict \
					 --main-module WASM --syntax-module WASM ~/wasm-semantics/.build/defn/wasm/wasm.k --directory ~/wasm-semantics/.build/defn/wasm \
		&& cd ~/wasm-semantics/.build/defn/wasm/wasm-kompiled \
[...]

PR #22, opened 2018-12-04
Fix for MacOS where absolute path messes up because of filename conventions

ehildenb:
@mrchico this should address #21

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 # Settings
 # --------
 
-build_dir:=$(CURDIR)/.build
+build_dir:=.build
 defn_dir:=$(build_dir)/defn
 k_submodule:=$(build_dir)/k
 pandoc_tangle_submodule:=$(build_dir)/pandoc-tangle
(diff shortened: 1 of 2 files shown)


MrChico:
Hmmm, the problem persists when I try this branch

MrChico:
Builds, but does not run. Am I missing some ocaml deps?
```
make test
./kwasm test tests/simple/arithmetic.wast
error loading shared library: dlopen(~/wasm-semantics/.build/defn/test/test-kompiled/realdef.cmxs, 10): Symbol not found: _caml_curry10
  Referenced from: ~/wasm-semantics/.build/defn/test/test-kompiled/realdef.cmxs
  Expected in: flat namespace
 in ~/wasm-semantics/.build/defn/test/test-kompiled/realdef.cmxsFatal error: exception Failure("could not load semantics")
[Error] Critical: Failed to execute program in ocaml. See output for error
information.
make: *** [tests/simple/arithmetic.wast.test] Error 113
```

ehildenb:
It's going to be hard for me to figure this out without access to your system. Can you look at target `ocaml-deps` in the `Makefile` of `evm-semantics` and start to pull over packages installed on the last line into this `Makefile` one at a time to the `wasm-semantics` makefile in target `ocaml-deps`? Then run `make ocaml-deps ; make build ; make test`, and if it succeeds with some of those packages added then we can add it to the makefile for wasm.

ehildenb:
This is fully subsumed by #38

PR #25, opened 2019-01-28
Makefile: java-deps, pandoc-deps and build-java targets

MrChico:
Includes https://github.com/kframework/wasm-semantics/pull/22.

Testing it locally it fails with:
```
     <funcs>
-      .FuncDefCellMap
+      .Map
     </funcs>
```
which is a problem I also encountered with my other PRs – there might be a discrepancy in how the different backends interpret empty cells here.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,11 @@ clean:
 # Build Dependencies (K Submodule)
 # --------------------------------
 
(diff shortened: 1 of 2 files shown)


hjorthjort:
Verifying: I can build and run using these changes.

ehildenb:
This has since been subsumed by #38. I have another PR in the pipeline which adds the Haskell and Java backend, but since the OCaml backend is working it's not urgent anymore.

Issue #26, opened 2019-01-28
Build issue: `opam switch 4.03.0+k` not working


hjorthjort:
Running `make deps` in clean repository. Also tried `make clean && make deps -B`` with same result (just a bit more output before).

Output of `make deps`:  
```                                                                                                 
opam init --quiet --no-setup                                                                                  
opam repository add k "/home/hjort/code/wasm/wasm-semantics/.build/k/k-distribution/target/release/k/lib/opam$
 \                                                                                                            
        || opam repository set-url k "/home/hjort/code/wasm/wasm-semantics/.build/k/k-distribution/target/rel$
ase/k/lib/opam"                                                                                            
[k] synchronised from file:///home/hjort/code/wasm/wasm-semantics/.build/k/k-distribution/target/release/k/li$
/opam                                                                                                         
[WARNING] The repository 'k' at                                                                               
          file:///home/hjort/code/wasm/wasm-semantics/.build/k/k-distribution/target/release/k/lib/opam       
          doesn't have a 'repo' file, and might not be compatible with this version of opam.                  
[NOTE] Repository at                                                                                          
       file:///home/hjort/code/wasm/wasm-semantics/.build/k/k-distribution/target/release/k/lib/opam doesn't  
       define its version, assuming it's 1.2.                                                                 
                                                                                                              
<><> Upgrading repositories from older opam format ><><><><><><><><><><><><><><>                              
Upgrading repository "k"...                                                                                   
[NOTE] Repository k has been added to the selections of switch 4.03.1+k only.                                 
       Run `opam repository add k --all-switches|--set-default' to use it in all existing switches, or in     
       newly created switches, respectively.                                                                  
                                                                                                              
opam update                                                                                                   
                                                                                                              
<><> Updating package repositories ><><><><><><><><><><><><><><><><><><><><><><>                              
[k] synchronised from file:///home/hjort/code/wasm/wasm-semantics/.build/k/k-distribution/target/release/k/li$
/opam                                                                                                         
[WARNING] The repository 'k' at                                                                               
          file:///home/hjort/code/wasm/wasm-semantics/.build/k/k-distribution/target/release/k/lib/opam       
          doesn't have a 'repo' file, and might not be compatible with this version of opam.                  
[default] no changes from https://opam.ocaml.org                                                              
[NOTE] Repository at                                                                                          
       file:///home/hjort/code/wasm/wasm-semantics/.build/k/k-distribution/target/release/k/lib/opam doesn't  
[...]

PR #27, opened 2019-01-28
Switch to CI based on jenkinsfile-dockerfile

ehildenb:


ehildenb:
@hjorthjort can you try out the new install instructions included with this one and see if it works on your system? In particular, the OCaml backend should work now if you follow the build instructions in the README. `make deps` alone is not enough anymore, we have the user install the ocaml dependencies manually because it installs things outside of the working directory.

Make sure that when you run `git submodule update --init --recursive` that you have checked out this current branch. Also, once you've run these ocaml setup instructions, you **shouldn't** have to run them again for any other K repo (at least, until we bump the OCaml version we use, which probably is at least 1 year out).

Also, @mrchico, can you try?

ehildenb:
Note that this PR also switches to our new Jenkinsfile/Dockerfile based build. It's for our internal CI setup.

Makefile
```diff
@@ -30,21 +30,16 @@ deps: $(k_submodule)/make.timestamp $(pandoc_tangle_submodule)/make.timestamp oc
...
-		&& mvn package -q -DskipTests
+		&& mvn package -q -DskipTests -Dllvm.backend.skip
 	touch $(k_submodule)/make.timestamp
 
 $(pandoc_tangle_submodule)/make.timestamp:
 	git submodule update --init -- $(pandoc_tangle_submodule)
 	touch $(pandoc_tangle_submodule)/make.timestamp
 
 ocaml-deps:
-	opam init --quiet --no-setup
```
dwightguth: See my comment on the evm pr

README.md
```diff
@@ -22,15 +22,28 @@ The following are needed for building/running KWASM:
 On Ubuntu >= 15.04 (for example):
 
 ```sh
-sudo apt-get install make gcc maven openjdk-8-jdk flex opam pkg-config libmpfr-dev autoconf libtool pandoc zlib1g-dev
+sudo apt install                                                         \
+        autoconf curl flex gcc libffi-dev libmpfr-dev libtool make maven \
+        opam openjdk-8-jdk pandoc pkg-config python3 python-pygments     \
```
dwightguth: I'm very skeptical of this list of dependencies also, because it includes some things from the other list that surely are either extraneous or refer only to the jello paper. I really feel strongly that this needs to be better tested or we're going to give users a bad install experience

hjorthjort:
> @hjorthjort can you try out the new install instructions included with this one and see if it works on your system? In particular, the OCaml backend should work now if you follow the build instructions in the README. `make deps` alone is not enough anymore, we have the user install the ocaml dependencies manually because it installs things outside of the working directory.
> 
> Make sure that when you run `git submodule update --init --recursive` that you have checked out this current branch. Also, once you've run these ocaml setup instructions, you **shouldn't** have to run them again for any other K repo (at least, until we bump the OCaml version we use, which probably is at least 1 year out).
> 
> Also, @MrChico, can you try?

Can confirm, works for me know.

Fixes #26

README.md
ehildenb: I would prefer to keep the Dockerfile's having the same prefix, since it will mostly be used on our CI servers.

README.md
```diff
@@ -22,15 +22,28 @@ The following are needed for building/running KWASM:
...
+```sh
+git submodule update --init --recursive
+./.build/k/k-distribution/src/main/scripts/bin/k-configure-opam-dev
+```
+
+**CAUTION**: It may prove useful to first do `rm -rf ~/.opam` if you've setup K projcets in the past and are experiencing trouble with the newest opam libraries.
+             This is a fairly destructive operation, and will break any other projects that depend on specific locally installed ocaml packages.
+
 Installing/Building
 -------------------
```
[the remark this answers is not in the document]
ehildenb: Fixed.

PR #28, opened 2019-01-30
Add `unreachable` instruction

hjorthjort:
I noticed this was missing in the control flow instruction set. [It is used a few times in the official test suite](https://github.com/search?q=unreachable+repo%3AWebAssembly%2Fspec+path%3A%2Ftest%2Fcore&type=Code), whereas the `trap` is not used in any programs as far as I can tell.

From page 69 in official spec:

`unreachable`
1. Trap.
unreachable ˓→ trap

### DIFF
--- a/wasm.md
+++ b/wasm.md
@@ -258,7 +258,7 @@ Structured Control Flow
 
 ```k
     syntax Instr ::= "unreachable"
- // -----------------------
+ // ------------------------------
     rule unreachable => trap
 ```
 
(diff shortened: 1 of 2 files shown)


wasm.md
```diff
@@ -254,6 +254,14 @@ Structured Control Flow
     rule <k> nop => . ... </k>
 ```
 
+`unreachable` causes an immediate `trap`.
+
+```k
+    syntax Instr ::= "unreachable"
+ // -----------------------
```
ehildenb: Please line the `---` underneath up with the end of the line above.

Sorry, I'm a stickler for formatting though :P

hjorthjort: Nothing to apologize for, I appreciate well though-out styling.

PR #38, opened 2019-02-06
Simplify Makefile, add instrumentation to kwasm script

ehildenb:
Hey @hjorthjort please test this locally (maybe also @cdetrio) and let me know/approve if it works for you and still passes tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,6 @@
 
 *.pdf
 *.sty
+
+.krun*
+.kompile*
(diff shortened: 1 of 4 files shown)


hjorthjort:
Builds and tests pass, but only after I install `time` as a dependency.
I use zsh normally, which has `time` built in. Seems Arch Linux doesn't ship with it installed as default.

Anyway, I would add it to the dependency list, just to be sure. The failure output was
```
which: no time in ([My entire $PATH variable as a string])
```

PR #41, opened 2019-02-11
Simplifying parsing for introduction of folded syntax by requiring paranthesis

ehildenb:
Hey @hjorthjort this prepares for introducing a folded syntax which works with the parser by forcing the use of paranthesis around the concrete syntax. Once it's applied to an argument, the parens go away (so that the semantics are given over the abstract non-paren form).

I also factored out an operator category `ConvOp`. Not sure how I feel about it, but I'll let you decide. It avoids the parsing issues with `_` by building it into the larger tokens (as you did), but still has it type-parametric on the output type (like the other opcodes). I could take it or leave it.

### DIFF
--- a/tests/simple/arithmetic.wast
+++ b/tests/simple/arithmetic.wast
@@ -1,151 +1,151 @@
-i32.const 5
-i32.const 7
-i32.add
+(i32.const 5)
+(i32.const 7)
+(i32.add)
 #assertTopStack < i32 > 12 "add"
 
-i32.const 5
-i32.const 7
-i32.sub
+(i32.const 5)
+(i32.const 7)
+(i32.sub)
 #assertTopStack < i32 > 2 "sub"
 
-i32.const 15
-i32.const 3
-i32.mul
+(i32.const 15)
+(i32.const 3)
+(i32.mul)
 #assertTopStack < i32 > 45 "mul"
 
-i32.const 3
-i32.const 15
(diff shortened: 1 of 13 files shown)


ehildenb:
@hjorthjort I've added `trap` cases for when we have type mis-matches for the return type of a `ConvOp`. I think actually, the input type shouldn't be checked, since it's checked at validation and the rest of the semantics just halts on type-failure (so I want to stick with that convention for consistency).

hjorthjort:
> @hjorthjort I've added `trap` cases for when we have type mis-matches for the return type of a `ConvOp`. I think actually, the input type shouldn't be checked, since it's checked at validation and the rest of the semantics just halts on type-failure (so I want to stick with that convention for consistency).

The issue with that is that it means that an instruction like `i64.wrap_i64` still parses, even though it is not in the instruction set. I guess that could be considered something the validator would catch, but I'm not a big fan of allowing "phantom instructions" in in at all.

hjorthjort:
I think the conversion operators make for an interesting design case for how to make K play along with the Wasm spec. It would be really cool if we could get close to the spec on this one, maybe even as a pedagogical example for the talk, as it could be a good showcase of what K can do: higher order functions, parametricity, abstraction.

ehildenb:
Lets punt on it being too permissive for now, and fix it if it fails a test in the validation test-suite.

PR #46, opened 2019-02-18
Operator abstractions

hjorthjort:
Creates separate abstractions for testing and comparison operators. Also groups these abstractions in the same place in the file as `UnOp` and `BinOp`. In that vein, it also moves `ConvOp` to the same place. 

Finally, I took some liberties with cleaning up the conversion operator: I would prefer not to trap on non-existent instructions. Traps seem like they should only happen when the spec says so, not for non-existent instructions.

### DIFF
--- a/wasm.md
+++ b/wasm.md
@@ -120,6 +120,34 @@ When a binary operator is the next instruction, the two arguments are loaded fro
          <stack> < ITYPE > C2 : < ITYPE > C1 : STACK => STACK </stack>
 ```
 
+### Test Operations
+
+Test operations consume one operand and produce a bool, which is an `i32` value.
+They are parametric in the operand type.
+
+```k
+    syntax Instr  ::= "(" IValType "." ITestOp ")"
+    syntax Int    ::=  #itestop(ITestOp, IValType, Int) [function]
+ // --------------------------------------------------------------
+    rule <k> ( ITYPE . TOP:ITestOp ) => . ... </k>
+         <stack> < ITYPE > SI1 : STACK => < i32 > #itestop(TOP, ITYPE, SI1) : STACK </stack>
+```
+
+### Comparison Operations
+
+Comparisons consume two operands and produce a bool, which is an `i32` value.
+They are parametric in the operand type.
+
+```k
+    syntax Instr  ::= "(" IValType "." RelOp ")"
+    syntax RelOp  ::= IRelOp
+ //                 | FRelOp
+    syntax Number ::= #relop(RelOp, IValType, Int, Int) [function]
+ // --------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


wasm.md
```diff
@@ -120,6 +120,59 @@ When a binary operator is the next instruction, the two arguments are loaded fro
...
+They are parametric in the operand type.
+For each element added to `RelOp`, function `#relOp` must be defined over it.
+Since comparison operations can not trap, it suffices to lift the values into a function.
+
+
+```k
+    syntax Instr  ::= "(" IValType "." RelOp ")"
+    syntax RelOp  ::= IRelOp
+ //                 | FRelOp
+    syntax Number ::= #relOp(RelOp, IValType, Int, Int) [function]
```
ehildenb: This will always return an `Int` right?

ehildenb:
I think we should be consistent about the syntax we use. For example, we use `IValType "." IBinOp Int Int` for the functional form of the binary operator, so I would imagine `IValType "." IRelOp Int Int` for the `RelOp`, for example. Similarly for the `ITestOp`.

I would also prefer the switch to using `function` to be in another PR (if at all). Right now we don't use `function` for the others so that we can uniformly handle the `undefined` value (before it ever makes it to the `<stack>` cell).

So I would imagine it looking more like:

```
syntax Instr ::= "(" IValType "." IRelOp ")" | IValType "." IRelOp Int Int
```

with rules that look very similar to the `BinOp` rules.

ehildenb:
Note that deviating from the abstract syntax of the spec seems perfectly acceptable to me, even if we maintain a 1-to-1 mapping between constructs (eg. `IValType "." IRelOp Int Int` <-> `#iRelOp(IValType, IRelOp, Int, Int)`).

wasm.md
hjorthjort: Yep, my bad. It doesn't matter now though, `#relOp` is removed.

hjorthjort:
Rebased onto master

wasm.md
```diff
@@ -163,35 +215,39 @@ Note that we do not need to call `#chop` on the results here.
...
-All of the following opcodes are liftings of the K builtin operators using the helper `#bool`.
+`eqz` checks wether its operand is 0.
 
 ```k
-    syntax IUnOp ::= "eqz"
- // ----------------------
-    rule <k> ITYPE . eqz I1 => < ITYPE > #bool(I1 ==Int 0) ... </k>
+    syntax ITestOp ::= "eqz"
+ // ------------------------
+    rule <k> ITYPE . eqz I => < i32 > #bool(I ==Int 0) ... </k>
```
ehildenb: Can we ignore the type here as well (ie. `_ . eqz I`) like you do below?

hjorthjort: Good catch, fixed.

Issue #34, opened 2019-02-05
Allow folded syntax


hjorthjort:
The following sequences of instructions should have the same semantics in Wasm:

```
;; 1
i32.const 5
i32.const 7
i32.add

;; 2
i32.add (i32.const 5) (i32.const 7)
```

However, only the first syntax is currently allowed, as far as I can tell. The second program halts at

```
 i32 . add ~> i32 . const 5 ~> i32 . const 7
```
with no matching rewrite rule.

Currently, parentheses are only used as a grouping mechanism, by the `[bracket]` annotation. From playing around with the Wasm reference interpreter, it seems to me that parentheses are not grouping, but actually alter the semantics. For example, this is valid:
```

i32.add (i32.const 5) (i32.const 7)
```

wheras this isn't:
```
i32.add ((i32.const 5)) ((i32.const 7))
```

Perhaps the handling of parentheses needs to be re-evaluated. Alternatively, an unfolder would be helpful, that could normalize the syntax to unfolded form, before execution.

hjorthjort:
I'd say that this has been addressed; we now (mostly) allow folded syntax in our rules.

PR #47, opened 2019-02-19
Enable folded syntax, several fixes to current operator syntax

ehildenb:


hjorthjort:
This closes #34

ehildenb:
@hjorthjort I've added tests that the `ite` operator is lazy in both branches, but strict in the condition. I've also added tests that `select` is strict in condition and both branches.

To do this, I needed to change `test.md` up a bit. That file should be re-reviewed. `wasm.md` didn't change again, except that I now force parens around `unrechable`.

test.md
```diff
@@ -16,13 +16,24 @@ Assertions
 These assertions will check the supplied property, and then clear that state from the configuration.
 In this way, tests can be written as a serious of setup, execute, assert cycles which leaves the configuration empty on success.
 
+We'll make `Assertion` a subsort of `Instr`, so that we can consume them with `trap` statements.
+
```
hjorthjort: If you want you can add a short sentence on why we want to consume traps the way we do here, took me a bit of thinking to see that it's needed for the select tests.

ehildenb: I added some more documentation of this.

PR #50, opened 2019-02-20
Separate lemmas into own file for verification, updates to kwasm script

ehildenb:
@hjorthjort this PR introduces the file `kwasm-lemmas.md`, where you can place facts about WASM simplification which probably shouldn't be directly in the semantics.

Note that we have few semantics which we use for commercial verification (KEVM being one of them), so it's not entirely clear where the "correct" place to put these lemmas is (whether in the definition or not).

Also, we add a couple more proofs to `simple-arithmetic-spec.k`. Feel free to add more if you feel adventurous.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-KWASM: Semantics of WebAssembly in K
+KWasm: Semantics of WebAssembly in K
 ====================================
 
 This repository presents a prototype formal semantics of [WebAssembly].
@@ -14,7 +14,7 @@ This repository generates the build-products for both backends in `.build/defn/j
 
 ### System Dependencies
 
-The following are needed for building/running KWASM:
+The following are needed for building/running KWasm:
 
 -   [git](https://git-scm.com/)
 -   [Pandoc >= 1.17](https://pandoc.org) is used to generate the `*.k` files from the `*.md` files.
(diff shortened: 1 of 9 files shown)


hjorthjort:
> Note that we have few semantics which we use for commercial verification (KEVM being one of them), so it's not entirely clear where the "correct" place to put these lemmas is (whether in the definition or not).

What's the connection between the semantics being used commercially and them being in one file or the other?

ehildenb:
I just mean to say that verification is very new (as in we only have a few semantics where any verification has happened, and fewer where it has happened at a large scale), so we haven't fully figured out how to structure such projects.

README.md
```diff
@@ -14,7 +14,7 @@ This repository generates the build-products for both backends in `.build/defn/j
 
 ### System Dependencies
 
-The following are needed for building/running KEVM:
+The following are needed for building/running KWASM:
```
hjorthjort: We should maybe start opting for "KWasm" or similar. "Wasm" isn't an acronym and therefore not capitalized, so the official abbreviation is "Wasm".

kwasm-lemmas.md
```diff
@@ -0,0 +1,21 @@
...
+module KWASM-LEMMAS
+    imports WASM
+```
+
+When reasoning about `#chop`, it's often the case that the precondition to the proof contains the information needed to indicate no overflow.
+In this case, it's simpler (and safe) to simply discard the `#chop`, instead of evaluating it.
+
+```k
+    rule #chop(< ITYPE:IValType > N) => < ITYPE > N requires N <Int #pow(ITYPE)
+```
```
hjorthjort: We should probably add `N >=Int 0` as well. It should be an invariant of `IValType`, but we might as well play defensively in the trusted base.

README.md
ehildenb: KWASM => KWasm, and WASM => WebAssembly throughout the repository.

This debate has come up from time to time on the ewasm team, it seems that WebAssembly is the only non-controversial choice. For KWasm I figure we can pick whatever name we like, but probably will be less controversial with KWasm.

kwasm-lemmas.md
ehildenb: Done.

PR #52, opened 2019-02-25
Generate golden files instead of updating them manually

hjorthjort:
When fiddling with the configuration, it becomes tedious to update the golden files after each change when you want to test. Here's a suggested way to make life easier.

I would also be happy to just keep this change in a script outside the repo on my end if we prefer having the golden files version-controlled.


ehildenb:
Lets keep this outside the repo. It deviates from all our other semantics build setups, and the configuration doesn't change often enough over the lifetime of a semantics to risk messing up test output because we aren't forced to manually update it when needed.

PR #53, opened 2019-02-27
Build using python/ninja-based KNinja

nishantjr:
This replaces the `Makefile`

TODOs:

* [ ] `./build --help` should also display `ninja`'s help options.
* [x] Test should output colored diffs.
* [ ] build the presentations in the media directory yet


nishantjr:
@dwightguth I can see your comment, and that you requested changes, but cannot see what the requested changes are.

nishantjr:
@ehildenb I've finished all the changes you recommended https://github.com/nishantjr/kninja/issues/1 (except the `kwasm` runner script).

Lets use this PR to coordinate

dwightguth:
As I said, I don't have any specific changes I'm requesting, I simply don't want this merged without consensus that ninja is the correct build system to use for all our semantics at the company.

ehildenb:
`./build clean` doesn't seem to work.

`./build` seems to update individual submodules as a target, should we do it once at the beginning of the `./build` script and then assume it's been done from there out?

Output log looks like:

```
[1/26] run: wasm-java tests/simple/comparison.wasm
[2/26] Checking .build/tests/simple/comparison.wast.wasm-java-run
```

I'd like `checking:` (or `diff:`)

ehildenb:
If proofs fail in CI, I'd like to see the end states that failed in the CI log (perhaps using `git diff` against some file with `#True` in it?)

nishantjr:
@ehildenb Let me know if there's anything else needed.

Jenkinsfile
```diff
@@ -13,46 +12,110 @@ pipeline {
...
-            [ "$nprocs" -gt '4' ] && nprocs=4
-            make test-proof -j"$nprocs"
-          '''
+      stages {
+        stage('Build') {
+          steps {
+            ansiColor('xterm') {
+              checkout scm
+              sh '''
+                ./build --opamroot ~/.opam wasm-java wasm-ocaml wasm-haskell
```
dwightguth: why does the user need to specify this --opamroot flag?

ehildenb: Default is `.build/opam` (inside the repo), which is nice for user builds so that we get local opam roots.

But for CI server, we want `$HOME/.opam` instead, so we can take advantage of dockerfile cacheing.

dwightguth: I'm not really sure why you would ever want a per-repository opam root for the ocaml backend.

ehildenb: For people who only install one semantics and otherwise don't care about K and have other OCaml projects which they don't want messed with. It localizes all changes to the repo, which is a clear win.

nishantjr: It is not required. I prefer that builds of a repo don't affect the users state outside of that repo, so KNinja defaults to setting `$OPAMROOT` to `.build/opam`. This means that users don't need to worry about K projects interfering with e.g. `coq` installs that need another `ocamlc` version. For the CI, we set it to `~/.opam` so that opam dependencies can be installed by the `Dockefile` saving build time.

dwightguth: Oh I see what you're saying. The real solution to that problem is to invoke eval `opam config env --switch 4.06.1+k` in the shell that executes ocaml commands, and then simply don't set our custom switch as the default when we create it.

dwightguth: Building an entire opam switch takes significant time and I have no intention of building it once for every project we build. You can do something cheap for now if you want, but it needs to be addressed before this kninja stuff could possibly be ready to be used as anything other than an experiment.

README.md
```diff
@@ -63,11 +63,10 @@ If you haven't already setup K's OCaml dependencies more recently than February
 Finally, you can install repository specific dependencies and build the semantics:
 
 ```sh
-make deps
-make build
+./build
 ```
 
-To only build specific backends, you can do `make build-java`, `make build-ocaml`, or `make build-haskell`.
+To only build specific backends, you can do `./build wasm-java`, `./build wasm-ocaml`, or `./build wasm-haskell`.
```
hjorthjort: How about the tests? Should they also be built using `ninja`? I assume we wouldn't want to have *both* a Makefile and the ninja system.

hjorthjort:
Two things about the test output: We should probably diff the files in the order we were doing previously. I think it makes more sense to see expected output as "removed" in the diff, and actual output as "added". Also, I don't get colored output from ninja, which I would like to help focus/declutter.

With ninja:

DeepinScreenshot_termite_20190402100339

With make:

DeepinScreenshot_termite_20190402101552

README.md
ehildenb: We were planning on having both systems for a little while, to make sure that we are not losing any features. I think in a few weeks, I would move the `Makefile` to the `media` directory and have it only for building the media targets (which currently KNinja does not cover).

The ninja build system is capturing dependencies more accurately, so running `./build test-exec-ocaml` will always check (1) ocaml deps installed, (2) ocaml version of semantics built, and then (3) run the tests.

nishantjr:
* Parallelism: Ninja too has a `-j` option, but this defaults to using as many cores as you have. Just run  `./build -j3`. I realize now, that adding option parsing to `KNinja` (for `--opamroot`) hides ninja's help message. I'll work on adding that. For now, you can run `ninja --help` to see the available options. These can be passed directly to `./build`
* Caching? I'm not sure whats causing this. Is it possible that some other dependencies were being built in the first call to `./build`?
* Colour: I'll work on adding this.
* > I'm a fan of moderate verbosity.

  Yes, I find this annoying too. I pipe ninja through `cat` or use the `-v` option. Unfortunately, ninja does not offer much in the way of customizing output, so I don't think there's anything I can do here.

* [ ] `./build --help` should also display `ninja`'s help options.
* [ ] Test should output colored diffs.

nishantjr:
Pushed a fix for the colored diff. Will be a bit before I can work on the help message @hjorthjort

Jenkinsfile
ehildenb: We are no longer using OPAMROOT, but setting `OPAMSWITCH = 4.06.1+k` as default switch in the upstream KNinja repo.

PR #55, opened 2019-03-07
Bugfix: first argument to `memory` is always the initial size

hjorthjort:
I found this as I was working on memories. Seems I didn't pay close enough attention when implementing the `(memory X)` instruction. Old implementation treated it as the max size, which is incorrect, it should be the initial size.

I also found some sensible validity checks were missing (not allowing any initial size or max size), so I added those.

### DIFF
--- a/wasm.md
+++ b/wasm.md
@@ -660,15 +660,15 @@ Currently, only one memory may be accessible to a module, and thus the `<memAddr
 **TODO**: There are many more valid ways to instantiate memory. Allow specifying just max or neither min or max, folded syntax, identifier, inline instantiation, and inline export and import.
 
 ```k
-    syntax Instr ::= "(" "memory"               ")"
-                   | "(" "memory" Int           ")" // Size only
-                   | "(" "memory" Int Int       ")" // Min and max.
+    syntax Instr ::= "(" "memory"                  ")"
+                   | "(" "memory"     Int          ")" // Size only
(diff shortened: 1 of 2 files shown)


wasm.md
```diff
@@ -661,13 +661,16 @@ Currently, only one memory may be accessible to a module, and thus the `<memAddr
...
                    | "(" "memory" Int Int       ")" // Min and max.
                    |     "memory" "{" Int MemBound "}"
  // --------------------------------------------------
-    rule <k> ( memory                 ) => memory { 0   .MemBound } ... </k>
-    rule <k> ( memory         MAX:Int ) => memory { 0   MAX       } ... </k>
-    rule <k> ( memory MIN:Int MAX:Int ) => memory { MIN MAX       } ... </k>
+    rule <k> ( memory                 ) => memory { 0                 .MemBound         }         ... </k>
+    rule <k> ( memory MIN:Int         ) => memory { MIN               .MemBound         }         ... </k>
+      requires MIN <=Int #maxMemorySize()
+    rule <k> ( memory MIN:Int MAX:Int ) => memory { MIN               MAX               }         ... </k>
```
ehildenb: Why is there so much extra horizontal whitespace on the RHS of these rules? Can we squish it down a bit?

hjorthjort: Oh, right, it's due to the rule coming underneath it (part of next PR), which is pretty wide. I'll fix it, just need to rebase a little.

PR #58, opened 2019-03-19
Add support for KLab in `kwasm` runner script.

ehildenb:
@hjorthjort if you could test that `./kwasm klab-run tests/simple/arithmetic.wast` and `./kwasm klab-prove tests/proofs/simple-arithmetic-spec.k` work if you put `klab/bin` on your PATH externally, I would appreciate it. If so, let's get this merged.

Remember you need to use `rad-rewrite` branch of KLab.

@mhhf you may be interested as well.

### DIFF
--- a/wasm.md
+++ b/wasm.md
@@ -64,7 +64,8 @@ WebAssembly instructions are space-separated lists of instructions.
     syntax Instrs ::= List{Instr, ""} [klabel(listInstr)]
  // -----------------------------------------------------
     rule <k> .Instrs           => .       ... </k>
-    rule <k> I:Instr IS:Instrs => I ~> IS ... </k>
+    rule <k> I:Instr .Instrs   => I       ... </k>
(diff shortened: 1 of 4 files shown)


wasm.md
```diff
@@ -64,7 +64,8 @@ WebAssembly instructions are space-separated lists of instructions.
     syntax Instrs ::= List{Instr, ""} [klabel(listInstr)]
  // -----------------------------------------------------
     rule <k> .Instrs           => .       ... </k>
-    rule <k> I:Instr IS:Instrs => I ~> IS ... </k>
+    rule <k> I:Instr .Instrs   => I       ... </k>
```
hjorthjort: What's the point of (re)introducing these rules?

ehildenb: Having the two rules (one which explicitly handles the `.Instrs` case) means that we won't generate a bunch of extra `.Instrs` in the tail of the instruction stack.

PR #60, opened 2019-03-28
Chalmers talk and report

hjorthjort:
In addition to introducing the report and presentation, this PR adds some more complex logic for building reports and presentations and restructures the media directory.


media/201903-report-chalmers.md
```diff
@@ -0,0 +1,750 @@
...
+A transition rule takes the form
+
+```k
+<cellA> X => Y </cellA>
+<cellB> Z => W </cellB>
+etc.
+```
+
+where the arrow represents a rewrite: the value to the left is rewritten to the
+value on the right. The above rule says that "if `<cellA>` contains `X` and `<cellB>` contains `Z`, change the contents of `<cellA>` to `Y` and the contents of `<cellB>` to `W`."
```
ehildenb: Emphasize that the rewrite on both cells happens simultaneously, and that the rewrite being contained inside the cell is only a syntactic convenience.

media/201903-report-chalmers.md
```diff
@@ -0,0 +1,750 @@
...
+variable points to and replacing what is there by the right-hand side of the
+assignment.
+
+```k
+    rule <k> X := I:Int => . ... </k>
+         <env>   ...  X |-> SX       ... </env>
+         <store> ... SX |-> (V => I) ... </store>
+```
+
+Note that, for this rule to apply, the right-hand side must be an integer literal, i.e., a fully evaluated expression.
```
ehildenb: You may need more explanation of this builtin map sort. Maybe even consider doing as you did above for the associative operator, by expanding it out to show that this actually is:

```
<env> MAP_HEAD   X |-> SX   MAP_TAIL </env>
```

and that the `...` is builtin K syntax which can be used on both sides of an `assoc comm` operator (or just on the tail side of an `assoc` operator).

media/201903-report-chalmers.md
```diff
@@ -0,0 +1,750 @@
...
+
+We either must give all $n$ arguments in the folded way, or none.
+
+Modules
+---
+
+All Wasm code is organized in modules, much like some languages organize all
+code into classes. A module may declare functions, allocate and modify their own
+linear memory, maintain module-global variables, etc. All these assets are
+internal to the module by default, but the module mayexplicitly export them with
```
ehildenb: "mayexplicitly => may explicitly"

media/201903-report-chalmers.md
```diff
@@ -0,0 +1,750 @@
...
+In \K
+----
+
+This is the full definition of the `(memory.grow)` operation:
+
+```k
+    syntax Instr ::= "(" "memory.grow"       ")"
+                   | "(" "memory.grow" Instr ")"
+                   | "grow" Int
+ // ---------------------------------------------------------
```
ehildenb: formatting here is kinda weird

media/201903-report-chalmers.md
hjorthjort: Done

media/201903-report-chalmers.md
hjorthjort: Done.

media/201903-report-chalmers.md
hjorthjort: Done.

media/201903-report-chalmers.md
hjorthjort: Can it be used for any `assoc` or `assoc comm` operator? I tried playing around with it, and couldn't get it to work if I defined my own operator.

hjorthjort: Done

ehildenb: No, only the builtins unfortunately. We would like more general `...`, but it hasn't been high priority since rarely do people need anything other than `List` (monoid), `Set`, or `Map`.

PR #65, opened 2019-05-20
[draft] Use KNinja to implement the runner interface

nishantjr:


ehildenb:
We've removed KNinja for now, since updating both Makefile and KNinja was too much maintenance overhead.

PR #69, opened 2019-05-28
Update submodule dependencies

ehildenb:
PR generated by './update-deps/update-deps'


hjorthjort:
I could keep LGTMing these, but I don't really see the point, as I don't know what tool is generating them or how to check that they are sound. Are you requesting my review so that I check that they work on my machine before we merge?

ehildenb:
@hjorthjort it's a fully automatic update of the submodule dependencies, just to make sure we don't fall behind master K and have a massive update procedure pile up on us.

I would request my own review, but since it's my github credentials doing the opening of the PR, GitHub won't allow me to. So you don't even need to make sure it runs on your machine, you can just approve and if it passes we merge.

PR #70, opened 2019-05-28
Config fix globals

QianyangPeng:
1. added another sort: syntax Mut ::= ".Mut" | "const" | "var"
2. globals now moved into <mainStore>, and <moduleInst> now has a <globalAddrs> for global addresses.
3. "init_global" is moved to test.md because it is used only for test. Maybe "init_locals" should also be moved to test.md in the future.
4. definition for init_global now changed to "init_global" INDEX GADDR. The test code also changed accordingly.

### DIFF
--- a/test.md
+++ b/test.md
@@ -10,6 +10,27 @@ module WASM-TEST
     imports WASM
 ```
 
+Helper Functions
+----------------
+
+`init_global` is a helper function that helps us to declare a new global variable.
+
+```k
+    syntax Instr ::= "init_global" Int Int
+ // --------------------------------------
+    rule <k> init_global INDEX GADDR => . ... </k>
+         <globalAddrs> GADDRS => GADDRS [ INDEX <- GADDR ] </globalAddrs>
+         <globals>
+           ( .Bag =>
+             <globalInst>
+               <gAddr> GADDR </gAddr>
+               ...
+             </globalInst>
+           )
+           ...
+         </globals>
+```
+
 Assertions
 ----------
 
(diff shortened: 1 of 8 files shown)


test.md
```diff
@@ -10,6 +10,27 @@ module WASM-TEST
...
+           ( .Bag =>
+             <globalInst>
+               <gAddr> GADDR </gAddr>
+               ...
+             </globalInst>
+           )
+           ...
+         </globals>
+```
+
```
ehildenb: Move this closer to where the code for `#assertGlobal` is. In the same section.

You can give `Globals` their own section (and separate them from `Locals`).

tests/simple/variables.wast
```diff
@@ -19,8 +19,8 @@ init_locals < i32 > 0 : < i32 > 0 : < i32 > 0 : .Stack
 
 ;; Test globals
 
-init_global 0 < i32 > 0
-init_global 1 < i32 > 0
+init_global 0 0
+init_global 1 1
```
ehildenb: I think this should be something like:

```
init_global 0 0
(global.set 0 (i32.const 0))
init_global 1 1
(global.set 1 (i32.const 0))
```

QianyangPeng: added tests, implemented folded version and passed local check.

wasm.md
```diff
@@ -526,20 +533,29 @@ The `*_local` instructions are defined here.
 ### Globals
 
 ```k
-    syntax Instr ::= "init_global" Int Val
-                   | "(" "global.get" Int ")"
+    syntax Instr ::= "(" "global.get" Int ")"
                    | "(" "global.set" Int ")"
+                   | "(" "global.set" Int Instr ")"
  // -----------------------------------------
```
ehildenb: Line up all the `")"` and make the `-----` extend to the end of the line:

```
    syntax Instr ::= "(" "global.get" Int       ")"
                   | "(" "global.set" Int       ")"
                   | "(" "global.set" Int Instr ")"
 // -----------------------------------------------
```

QianyangPeng: fixed

PR #81, opened 2019-06-06
kwasm: updates pulled over from KEVM

ehildenb:
-   Separate out the step which actually launches KLab from `./kwasm klab-(run|prove)` into new command `./kwasm klab-view`, which allows asynchronously calling them.
-   Use the filename of the spec/program being proved to decide `--state-log-id`, so that many proofs can be run/viewed at once using KLab without overwriting each others results.
-   Add test `test-klab-prove` to make sure that `./kwasm klab-prove` isn't accidentally borked.
-   All testing is handled directly by `Makefile`, instead of relying on `./kwasm test`.
-   Eliminate an unneeded source of branching in the semantics due to bug in rule.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,24 +110,18 @@ Run the file `tests/simple/arithmetic.wast`:
 ./kwasm run tests/simple/arithmetic.wast
 ```
 
-Run the same file as a test:
-
-```sh
-./kwasm test tests/simple/arithmetic.wast
-```
-
-To run proofs, you can similarly use `./kwasm`.
+To run proofs, you can similarly use `./kwasm`, but most specify the module to use for proving.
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -110,24 +110,18 @@ Run the file `tests/simple/arithmetic.wast`:
...
 ```
 
-Run the same file as a test:
-
-```sh
-./kwasm test tests/simple/arithmetic.wast
-```
-
-To run proofs, you can similarly use `./kwasm`.
+To run proofs, you can similarly use `./kwasm`, but most specify the module to use for proving.
```
hjorthjort: most => must

ehildenb: Fixed!

ehildenb:
Will merge when tests pass.

PR #92, opened 2019-06-14
Data segments for initializing memory

hjorthjort:
I got sick of trying to get Wasm strings to work as intended, and the solution seemed unnecessary frail and complex to begin with. So I opted for a simpler version, that could either be dealt with later either in K or by pre-processing.

Basically, instead of dealing with arbitrary Unicode strings with escape sequences, I just said that `data` takes a list of bytes (`Ints`). A pre-processor would just go over the actual bytes and put them in as `Int`s, after checking for escape sequences.

This felt like the best solution for now, so we can move on, and consider fully compatible Wasm-strings a nice-to-have.

Hope you agree.


wasm.md
```diff
@@ -824,6 +827,8 @@ Currently, only one memory may be accessible to a module, and thus the `<memAddr
            )
            ...
          </mems>
+
+    syntax MemId ::= Identifier | Int
```
ehildenb: Sort `Index` declared in `data.md` can probably be used here instead.

wasm.md
```diff
@@ -1011,6 +1016,52 @@ The maximum of table size is 2^32 bytes.
...
+           <mAddr> ADDR </mAddr>
+           <mdata>   DATA
+                  => #clearRange(DATA, OFFSET, OFFSET +Int #dataStringsLength(STRING) -Int 1) [ OFFSET := #dataStrings2int(STRING)]
+           </mdata>
+           ...
+         </memInst>
+
+    syntax Int ::= #lengthDataPages        ( Data ) [function]
+                 | Int "/ceilInt" Int          [function]
+ // -----------------------------------------------------
```
ehildenb: Formatting here is kinda weird.

wasm.md
```diff
@@ -1011,6 +1016,52 @@ The maximum of table size is 2^32 bytes.
...
+
+    rule I1 /ceilInt I2 => (I1 /Int I2) +Int #if I1 modInt I2 ==Int 0 #then 0 #else 1 #fi
+```
+
+The `data` instruction takes an offset, which is an instruction.
+This is not optional.
+
+```k
+    syntax DataOffset ::= "(" "offset" Instr ")"
+                        | Instr
```
ehildenb: Hmmm, this looks weird to me. Why can `DataOffset` be an `Instr`?

wasm.md
ehildenb: Sorry, the sort to use is `TextFormatIdx`, declared in `data.md`.

wasm.md
hjorthjort: Yeah, it seems a bit redundant, but it's explicitly allowed (and occurs in the official tests):

> As an abbreviation, a single instruction may occur in place of the offset:
> instr ≡ ‘( ‘offset’ instr ‘)’

These are both official tests:
```
  (data (offset (i32.const 0)))
  (data (i32.const 0))
```

hjorthjort: In fact, that's true not just for `data`, but for `elem` as well, so it should probably just be named `offset` or something.

wasm.md
hjorthjort: Thanks! Fixed.

wasm.md
hjorthjort: Hopefully that's better. Also got rid of `#if` in favor of `requires`

wasm.md
```diff
@@ -934,16 +939,16 @@ The `offset` parameter is added to the the address given on the stack, resulting
 The `align` parameter is for optimization only and is not allowed to influence the semantics, so we ignore it.
 
 ```k
-    syntax MemArg ::= Offset | Align | Offset Align
-    syntax Offset ::= "offset" "=" Int
-    syntax Align  ::= "align"  "=" Int
- // ----------------------------------
+    syntax MemArg ::= OffsetArg | AlignArg | OffsetArg AlignArg
+    syntax OffsetArg ::= "offset" "=" Int
+    syntax AlignArg  ::= "align"  "=" Int
```
QianyangPeng: Maybe we can change it to "offset=" and "align=", because in the specs it says "no white space is allowed around the ‘=’".

wasm.md
```diff
@@ -1011,6 +1016,58 @@ The maximum of table size is 2^32 bytes.
...
+```k
+    syntax Instr ::= Data
+    syntax Data ::= "(" "data"     TextFormatIdx Offset DataStrings ")"
+                  | "(" "data"                   Offset DataStrings ")"
+                  |     "data" "{" TextFormatIdx        DataStrings "}"
+ // -------------------------------------------------------------------
+    // Default to memory 0.
+    rule <k> ( data       OFFSET STRINGS ) => ( data 0 OFFSET STRINGS ) ... </k>
+    rule <k> ( data MEMID    IS:Instr STRINGS ) => IS ~> data { MEMID STRINGS } ... </k>
+    rule <k> ( data MEMID (offset IS) STRINGS ) => IS ~> data { MEMID STRINGS } ... </k>
```
QianyangPeng: alignment is weird

wasm.md
hjorthjort: Oh, in that case sure. Might help the parser as well.

wasm.md
```diff
@@ -934,16 +939,16 @@ The `offset` parameter is added to the the address given on the stack, resulting
...
 
 ```k
-    syntax MemArg ::= Offset | Align | Offset Align
-    syntax Offset ::= "offset" "=" Int
-    syntax Align  ::= "align"  "=" Int
- // ----------------------------------
+    syntax MemArg ::= OffsetArg | AlignArg | OffsetArg AlignArg
+    syntax OffsetArg ::= "offset=" Int
+    syntax AlignArg  ::= "align="  Int
+ // -------------------------------------
```
ehildenb: formatting

wasm.md
```diff
@@ -1011,6 +1016,58 @@ The maximum of table size is 2^32 bytes.
...
+    syntax Int ::= #lengthDataPages ( DataStrings ) [function]
+ // ----------------------------------------------------------
+    rule #lengthDataPages(DS:DataStrings) => #dataStringsLength(DS) /ceilInt #pageSize()
+
+    syntax Int ::= Int "/ceilInt" Int [function]
+ // --------------------------------------------
+    rule I1 /ceilInt I2 => (I1 /Int I2)
+      requires I1 modInt I2  ==Int 0
+    rule I1 /ceilInt I2 => (I1 /Int I2) +Int 1
+      requires I1 modInt I2 =/=Int 0
```
ehildenb: let's do one rule/line here, line up the `requires` clauses.

ehildenb: Also I'm not in love with using `modInt` here, as it's historically been not so great when passed to Z3.

In KEVM, we have the following operator:

```
    syntax Int ::= Int "up/Int" Int [function] 
 // ------------------------------------------ 
    rule I1 up/Int 0  => 0 
    rule I1 up/Int 1  => I1 
    rule I1 up/Int I2 => (I1 +Int (I2 -Int 1)) /Int I2 requires I2 >Int 1 
```

I'll leave it up to you whether to change this, since I don't see it causing problems yet and `Data` strings should be concrete for the most part (since they are part of the input program).

hjorthjort: Good one. I'd rather reuse the KEVM one.

hjorthjort: Is the second rule of any use? It should be the exact same as the third when `I2 == 1`

PR #95, opened 2019-06-17
Categorizing declarations for a better program structure

QianyangPeng:
Now we define 4 types of `statements` instead of considering everything as `instruction`:
  - Instruction (`Instr`): Administrative or computational instructions. 
  - Definition  (`Defn`): The declarations of `type`, `func`, `table`, `mem` etc.
  - Auxiliary   (`Auxil`): Functions that only used in our KWASM semantics but not in the official specification including assertions, invoke a function by imported name, and maybe instantiate a module?
  - The Declaration of a module.

### DIFF
--- a/test.md
+++ b/test.md
@@ -16,17 +16,18 @@ Assertions
 These assertions will check the supplied property, and then clear that state from the configuration.
 In this way, tests can be written as a serious of setup, execute, assert cycles which leaves the configuration empty on success.
 
-We'll make `Assertion` a subsort of `Instr`, so that we can easily consume `Instr` with `trap`s, but not consume `Assertion`s.
+We'll make `Assertion` a subsort of `Auxil`, so that we can easily consume `Instr` with `trap`s, but not consume `Assertion`s.
 This will allow `trap` to "bubble up" (more correctly, to "consume the continuation") until it reaches its paired `#assertTrap_` statement.
 
 ```k
-    syntax Instr ::= Assertion
+    syntax Decl  ::= Auxil
+    syntax Auxil ::= Assertion
  // --------------------------
     rule <k> trap ~> (L:Label => .) ... </k>
     rule <k> trap ~> (.Instrs => .) ... </k>
-    rule <k> trap ~> (I:Instr => .) ... </k> requires notBool isAssertion(I)
+    rule <k> trap ~> (I:Instr => .) ... </k>
 
-    rule <k> trap ~> (I:Instr IS:Instrs => I ~> IS) ... </k>
+    rule <k> trap ~> (D:Decl DS:Decls => D ~> DS) ... </k>
 ```
 
 ### Trap Assertion
@@ -86,7 +87,7 @@ The operator `#assertLocal`/`#assertGlobal` operators perform a check for a loca
 `init_global` is a helper function that helps us to declare a new global variable.
 
 ```k
-    syntax Instr ::= "init_global" Int Int
+    syntax Auxil ::= "init_global" Int Int
  // --------------------------------------
     rule <k> init_global INDEX GADDR => . ... </k>
          <globalIndices> GADDRS => GADDRS [ INDEX <- GADDR ] </globalIndices>
@@ -216,7 +217,7 @@ Clear Module Instances
 The modules are cleaned all together after the test file is executed.
 
 ```k
-    syntax Instr ::= "#clearModules"
+    syntax Auxil ::= "#clearModules"
  // --------------------------------
     rule <k> #clearModules => . ... </k>
          <nextFreshId> _ => 0 </nextFreshId>
@@ -241,7 +242,7 @@ Function Invocation
 We allow to `invoke` a function by its exported name in the test code.
 
 ```k
-    syntax Instr ::= "(" "invoke" String ")"
+    syntax Auxil ::= "(" "invoke" String ")"
  // ----------------------------------------
     rule <k> ( invoke ENAME:String ) => ( call TFIDX ) ... </k>
          <exports> ... ENAME |-> TFIDX ... </exports>
--- a/test.md
+++ b/test.md
@@ -10,6 +10,16 @@ module WASM-TEST
     imports WASM
 ```
 
+Auxiliary
+---------
+
+Here we extend the sort `Stmt` by adding a new subsort called `Auxil`.
+This subsort contains Auxiliary functions that only used in our KWASM semantics but not in the official specification including assertions, initializing a global variable and the invocation of a function by exported name.
+
+```k
+    syntax Stmt ::= Auxil
+```
+
 Assertions
 ----------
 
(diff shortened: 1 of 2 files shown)


wasm.md
```diff
@@ -88,14 +88,30 @@ Instructions
 
 ### Sequencing
 
-WebAssembly instructions are space-separated lists of instructions.
+WebAssembly code consists of sequences of declarations (`Decls`).
+We define 4 types of declarations:
+  - Instruction (`Instr`): Administrative or computational instructions. 
+  - Allocation  (`Alloc`): The declarations of `type`, `func`, `table`, `mem` etc.
```
hjorthjort: The idea is excellent. But I think the names can be improved. This is a major entry point into this semantics file, and hopefully many people are going to use it, read it, debug it, study it, etc., so we should make sure we have the most helpful names possible. I think we can do better, and here's a short essay on the topic.

# Alloc

I don't think `Alloc` is a fitting name, although I think I see what you're going for. "Allocating" memories and tables makes sense, "allocating" a function seems a bit off but okay, seeing how we put it into the main store. But "allocating" an export is just weird. I also think the "allocation" really should happen during module instantiation, which we may want to make into a separate step later, as you mention in the next bullet.

It seems the word of choice in the spec is "definition", and sometimes "declaration". For example, section 2.5 states (emphasis mine):
> A module collects *definitions* for types, functions, tables, memories, and globals. In addition, it can *declare* imports and exports and provide initialization logic in the form of data and element segments or a start function.

Later on we find phrases like
>The data component of a module *defines* a vector of data segments that initialize a range of memory

> The start component of a module declares the function index of a start function

> The imports component of a module *defines* a set of imports

In section 6.6 I only see the word "define" being used for these top-level constructs in a module.

Either way, I think `Def` is the right name for what is now `Alloc`, since all the things in a module can be thought to be "defined":
* types
* funcs
* tables
* mems
* globals
* elem
* data
* start
* imports
* exports

We can think of each of these to be "defining" a module, which is to be later "instantiated".

# Decl

I also do not think it makes sense to talk about a KWasm file containing a list of "declarations" (`$PGM:Decls`), so `Decl` is also slightly off. For example, is `(i32.add (i32.const 1) (i32.const 2))` really a declaration? I think that sounds confusing. This is a place where we are deviating from the spec, and hopefully a lot of people will read and use our spec, so I think it matters to get this right.

The only other two options I can think of is
* `Expr`: Makes sense when looking at Wasm as a Lisp-like, S-expression based language. They are not *pure* expressions, since they may have side effects, but expressions nontheless.
* `Bare`: This keyword makes it explicit that this is what we, in KWasm, allow to appear as bare constructs directly in a file.

To complicate things further, we could either keep `Decl` as in the final use for `module`, and make `Decl` a subsort of `Expr`/`Bare`, which means we have an additional sort (additional confusion), but also an explicit production for what we allow as bare expressions; or, we could make `Expr`/`Bare` the construct for making modules, meaning we would just replace `Decl` with `Expr`/`Bare`.

With a change like this, `Decl` is only used for modules, so it may make more sense to name it `ModuleDecl` for readability.

# Suggestions

I favor using `Def` instead of `Alloc`, using `ModuleDecl` for modules, and making `Bare` the top-level construct. So below here, we would have productions like:

```k
    syntax Bare    ::= Instr | Def | ModuleDecl
    syntax Bares  ::= Instrs
                    | Defs
                    | ModuleDecls
                    | List{Bare, ""} [klabel(listBare)]
```

I think that's pretty readable and gives a good overview of what we allow. We should also have a sentence on that we are more permissive than Wasm, in that we do allow instructions and definitions at the top level.

## The other suggestions, in order of preference:

* `Instr`, `Def`, `Bare` 
* `Instr`, `Def`, `ModuleDecl`, `Expr` 
* `Instr`, `Def`, `Expr`

wasm.md
```diff
@@ -88,14 +88,30 @@ Instructions
...
+  - Auxiliary   (`Auxil`): Functions that only used in our KWASM semantics but not in the official specification including assertions, invoke a function by imported name, and maybe instantiate a module?
```
hjorthjort: Maybe put this line in `test.md`, at least until we use `Auxil` in this file?

wasm.md
```diff
@@ -661,9 +677,10 @@ A type use should start with `'(' 'type' x:typeidx ')'` followed by a group of i
 Type could be declared explicitly and could optionally bind with an identifier.
 
 ```k
-    syntax Instr ::= "(type" "(" "func" TypeDecls ")" ")"
-                   | "(type" Identifier "(" "func" TypeDecls ")" ")"
- // ----------------------------------------------------------------
+    syntax Alloc     ::= TypeAlloc
+    syntax TypeAlloc ::= "(type" "(" "func" TypeDecls ")" ")"
```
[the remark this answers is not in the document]
QianyangPeng: Yes. Making different subsort is used for reordering these `Alloc`s, which is not used yet for now.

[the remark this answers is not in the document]
hjorthjort: Ah, like when grabbing all the functions of a module, then all the mems, etc? That's a clever way to do it!

wasm.md
hjorthjort: New suggestion:

```k
    syntax Stmt   ::= Instr | Defn | ModuleDecl
    syntax Stmts  ::= Instrs
                    | Defs
                    | ModuleDecls
                    | List{Bare, ""} [klabel(listBare)]
```

ehildenb: So maybe we can try:

```
    syntax Stmt   ::= Instr | Defn | ModuleDecl
    syntax Instrs  ::= List{Instr, ""} [klabel(listInstr)]
    syntax Stmts  ::= List{Bare, ""} [klabel(listStmt)]

rule <k> IS:Instrs => asStatements(IS) ... </k>

syntax Stmts ::= asStatements ( Instrs ) [function]
rule asStatements(.Instrs) => .Stmts
rule asStatements(listInstr(I, IS)) => listStmt(I, asStatements(IS))
```

ehildenb: Might have to say `.listInstr` instead `.Instrs`.

test.md
```diff
@@ -16,17 +16,18 @@ Assertions
 These assertions will check the supplied property, and then clear that state from the configuration.
 In this way, tests can be written as a serious of setup, execute, assert cycles which leaves the configuration empty on success.
 
-We'll make `Assertion` a subsort of `Instr`, so that we can easily consume `Instr` with `trap`s, but not consume `Assertion`s.
+We'll make `Assertion` a subsort of `Auxil`, so that we can easily consume `Instr` with `trap`s, but not consume `Assertion`s.
```
hjorthjort: `Instr` should be changed to `Auxil`

wasm.md
hjorthjort: And no `ModuleDecl`, right? Just let `(module ...)` be of sort `Stmt`.

test.md
QianyangPeng: I think we are consuming `Instr` not `Auxil`?

hjorthjort: Sorry, I meant `Instr` and `Auxil` should be changed to `Stmt` or whatever our top level is, right? So a trap consumes everything that is top level but isn't an assertion. Or at least it should consume both `Instr` and `Defn`.

test.md
```diff
@@ -10,24 +10,34 @@ module WASM-TEST
...
+    syntax Auxil ::= Assertion
  // --------------------------
     rule <k> trap ~> (L:Label => .) ... </k>
     rule <k> trap ~> (F:Frame => .) ... </k>
     rule <k> trap ~> (.Instrs => .) ... </k>
-    rule <k> trap ~> (I:Instr => .) ... </k> requires notBool isAssertion(I)
+    rule <k> trap ~> (I:Instr => .) ... </k>
 
-    rule <k> trap ~> (I:Instr IS:Instrs => I ~> IS) ... </k>
+    rule <k> trap ~> (S:Stmt SS:Stmts => S ~> SS) ... </k>
```
hjorthjort: If a module is being defined and a `trap` occurs, the trap should propagate all the way to the top level. From the official test:
```
(assert_trap
  (module (func $main (unreachable)) (start $main))
  "unreachable"
)
```

test.md
ehildenb: Fixed.

PR #97, opened 2019-06-18
Start function

hjorthjort:
Low hanging fruit: Implement the `start` definition.

### DIFF
--- a/test.md
+++ b/test.md
@@ -23,6 +23,7 @@ This will allow `trap` to "bubble up" (more correctly, to "consume the continuat
     syntax Instr ::= Assertion
  // --------------------------
     rule <k> trap ~> (L:Label => .) ... </k>
+    rule <k> trap ~> (F:Frame => .) ... </k>
     rule <k> trap ~> (.Instrs => .) ... </k>
     rule <k> trap ~> (I:Instr => .) ... </k> requires notBool isAssertion(I)
 
--- a/wasm.md
+++ b/wasm.md
@@ -1151,6 +1151,17 @@ The `data` initializer simply puts these bytes into the specified memory, starti
     rule I1 up/Int I2 => (I1 +Int (I2 -Int 1)) /Int I2 requires I2 >Int 0
 ```
 
+Start Function
+--------------
+
+```k
+    syntax Instr ::= "(" "start" TextFormatIdx ")"
+ // ----------------------------------------------
+    rule <k> ( start TFIDX ) => ( invoke FADDR ) ... </k>
+         <funcIds> IDS </funcIds>
+         <funcIndices> ... #ContextLookup(IDS , TFIDX) |-> FADDR ... </funcIndices>
+```
+
 Module Declaration
 ------------------
 
--- a/tests/simple/start.wast
+++ b/tests/simple/start.wast
@@ -1,0 +1,28 @@
+(memory 1)
+(func $inc
+  (i32.store8
+    (i32.const 0)
+    (i32.add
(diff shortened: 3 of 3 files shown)


hjorthjort:
FYI: both tests are lifted almost verbatim from the reference tests, `start.wast`.

test.md
```diff
@@ -23,6 +23,7 @@ This will allow `trap` to "bubble up" (more correctly, to "consume the continuat
     syntax Instr ::= Assertion
  // --------------------------
     rule <k> trap ~> (L:Label => .) ... </k>
+    rule <k> trap ~> (F:Frame => .) ... </k>
     rule <k> trap ~> (.Instrs => .) ... </k>
     rule <k> trap ~> (I:Instr => .) ... </k> requires notBool isAssertion(I)
```
hjorthjort: As a side note relating to #95, perhaps all these belong in `wasm.md`? If we make `Assertion` a subsort of `Bare`, then we won't need the final rule's side condition: a `trap` simply eats the instructions, frames, labels, definitions, declarations, etc, on the stack. But it doesn't eat `Bare`, or `Auxil`.

QianyangPeng: I agree this should be in wasm.md

wasm.md
```diff
@@ -1151,6 +1151,17 @@ The `data` initializer simply puts these bytes into the specified memory, starti
...
 
+Start Function
+--------------
+
+```k
+    syntax Instr ::= "(" "start" TextFormatIdx ")"
+ // ----------------------------------------------
+    rule <k> ( start TFIDX ) => ( invoke FADDR ) ... </k>
+         <funcIds> IDS </funcIds>
+         <funcIndices> ... #ContextLookup(IDS , TFIDX) |-> FADDR ... </funcIndices>
```
QianyangPeng: This might be correct... But note that in the section 2.5 of specs, `start` should be stored in module instance. Maybe we can keep this implementation for now.

hjorthjort: Hmm, sort of. 2.5 specifies that it's stored in the *module*, but that refers to the module declaration, not the instance, if I understand correctly. But, I guess our semantics define the way a module is instantiated, right? Like, the `(module ... )` is a syntactic object, and the semantics we have instantiate it by allocating functions, memories, etc. And the `start` function is called at instantiation time.

Maybe a good way to do this is to, whenever `( module ... )` is encountered, store the entire module definition in an array or something, and also have an array of instances. Then "instantiating" is just grabbing the definition and putting it at the front of the K cell, which allocates everything and stores the instance.

```k
rule <k> M:ModuleDecl => . ... </k>
     <nextModuleDefIdx> NEXT => NEXT +Int 1 </nextModuleDefIdx>
     <moduleDefs> ... .Map => NEXT |-> M ... </moduleDefs>
rule <k> ( instantiate IDX ) => DEFS ... </k>
     <moduleDefs> ... IDX |-> ( module DEFS ) ... </moduleDefs>
```

If we do it like that, then I guess the instance wont be carrying around its `start` funtction: it just gets called, and the module gets its initial state after.

hjorthjort: If you look at the end of 4.5.3 ("Modules") it specifies the allocation function, which doesn't care about `start`, and then 4.5.4 describes instantiation, which basically calls the allocation function, and proceeds to do `elem`, `data` and `start`.

So one way to stay closer to the spec is to have an `(alloc MODULE_IDX)` instruction, and one `(instantiate MODULE_IDX)` instruction. It may not be necessary, but I guess it should make for easier implementation, since it splits up a big job into two parts.

PR #101, opened 2019-06-19
Disambiguate Stmts/Instrs in proof

hjorthjort:
I'm getting warnings about parser ambiguity from `simple-arithmetic-spec.k`. This fixes it. Slightly annoying to have to do, but I guess it's worth it for the separation of sorts.


ehildenb:
Fixed by #102

PR #103, opened 2019-06-19
Registering modules

### LINKED ISSUE #99: Supporting `import` and multiple modules.
Objective: Handle the test cases in wasm-semantics/tests/wasm-tests/test/core/import.wast

In official test cases, they support the functionality of "registering" a recently defined module, and only then the module could be used for imports. In other word, multiple module instances should be supported in order to implement `import`.

"register" gives module a name, so that `import` will know the module name to be imported from.

hjorthjort:
We will need this for imports.

I started working on adding more `Auxil` instructions from the official interpreter and started with `register`. This way of storing allocated modules is roughly what we discussed, IIRC.

One thing I want to discuss is whether we want to have a separate "embedder" module, which would imitate the reference interpreter, and have `test.md` be our own tests, that could optionally be discarded after we pass the official tests. Now I'm keeping them in the same file, which seems wrong. But that refactoring can be done at a later point.

I'd argue this fixes #99, because that issue is mostly about registering modules. We haven't started on imports yet, but I conjecture they would be pretty straightforward with this configuration.

### DIFF
--- a/test.md
+++ b/test.md
@@ -265,13 +265,15 @@ These assertions act on the last module defined.
 
 ```k
     syntax Assertion ::= "#assertUnnamedModule"
+    syntax Assertion ::= "#assertUnnamedModuleAux" Int
     syntax Assertion ::= "#assertNamedModule" Identifier
  // ----------------------------------------------------
-    rule <k> #assertUnnamedModule => . ... </k>
+    rule <k> #assertUnnamedModule => #assertUnnamedModuleAux NEXT ... </k>
          <nextModuleIdx> NEXT => NEXT -Int 1 </nextModuleIdx>
+    rule <k> #assertUnnamedModuleAux IDX => . ... </k>
          <moduleInstances>
            ( <moduleInst>
-               <modIdx> NEXT -Int 1 </modIdx>
+               <modIdx> IDX </modIdx>
                ...
              </moduleInst>
           => .Bag
--- a/test.md
+++ b/test.md
@@ -50,7 +50,7 @@ We will reference modules by name in imports.
  // --------------------------------------------------------
     rule <k> ( register S ) => ( register S (NEXT -Int 1) )... </k> // Register last instantiated module.
          <nextModuleIdx> NEXT </nextModuleIdx>
-         requires NEXT >Int 0
+      requires NEXT >Int 0
     rule <k> ( register S ID:Identifier ) => ( register S IDX ) ... </k>
          <moduleIds> ... ID |-> IDX ... </moduleIds>
     rule <k> ( register S:String IDX:Int ) => . ... </k>
--- a/test.md
+++ b/test.md
@@ -244,6 +244,10 @@ The modules are cleaned all together after the test file is executed.
            <globalIndices> _ => .Map </globalIndices>
            <exports>       _ => .Map </exports>
          </moduleInst>
+         <moduleIds> _ => .Map </moduleIds>
+         <nextModuleIdx> _ => 0 </nextModuleIdx>
+         <moduleInstances> _ => .Map </moduleInstances>
+         <moduleRegistry> _ => .Map </moduleRegistry>
 ```
 
 Official Test Assertions
@@ -265,6 +269,17 @@ We allow to `invoke` a function by its exported name in the test code.
 
 ### Registering Modules
 
+We first define how modules get stored away.
+
+```k
+    rule <k> #storeInstance => . ... </k>
+         <moduleInst> INST </moduleInst>
+         <nextModuleIdx> NEXT => NEXT +Int 1 </nextModuleIdx>
+         <moduleInstances> ... .Map => NEXT |-> INST ... </moduleInstances>
+```
+
+We will reference modules by name in imports.
+`register` is the instruction that allows us to associate a name with a module.
 
 ```k
     syntax Assertion ::= "(" "register" String            ")"
--- a/test.md
+++ b/test.md
@@ -246,8 +246,13 @@ The modules are cleaned all together after the test file is executed.
          </moduleInst>
 ```
 
-Function Invocation
--------------------
+Official Test Assertions
+------------------------
+
+The official test suite contains some special auxillary instructions outside of the standard Wasm semantics.
+The reference interpreter is a particular embedder with auxillary instructions, specified in [spec interpreter](https://github.com/WebAssembly/spec/blob/master/interpreter/README.md).
+
+### Function Invocation
 
 We allow to `invoke` a function by its exported name in the test code.
 
@@ -256,6 +261,24 @@ We allow to `invoke` a function by its exported name in the test code.
  // ----------------------------------------
     rule <k> ( invoke ENAME:String ) => ( call TFIDX ) ... </k>
          <exports> ... ENAME |-> TFIDX ... </exports>
+```
+
+### Registering Modules
+
+
+```k
+    syntax Assertion ::= "(" "register" String            ")"
+                       | "(" "register" String Identifier ")"
+                       |     "register" String Int
+ // ----------------------------------------------
+    rule <k> ( register S ) => register S (NEXT -Int 1) ... </k> // Register last module.
+         <nextModuleIdx> NEXT </nextModuleIdx>
+         requires NEXT >Int 0
+    rule <k> ( register S ID:Identifier ) => register S IDX ... </k>
+         <moduleIds> ... ID |-> IDX ... </moduleIds>
+    rule <k> register S:String IDX:Int => . ... </k>
+         <moduleInstances> ... IDX |-> INST ... </moduleInstances>
+         <moduleRegistry> ... .Map => S |-> INST ... </moduleRegistry>
 ```
 
 ```k
--- a/test.md
+++ b/test.md
@@ -118,7 +118,12 @@ The operator `#assertLocal`/`#assertGlobal` operators perform a check for a loca
          <locals> ... (INDEX |-> VALUE => .Map) ... </locals>
 
     rule <k> #assertGlobal INDEX VALUE _ => . ... </k>
-         <globalIndices> ... INDEX |-> GADDR ... </globalIndices>
+         <curModIdx> CUR </curModIdx>
+         <moduleInst>
+           <modIdx> CUR </modIdx>
+           <globalIndices> ... INDEX |-> GADDR ... </globalIndices>
+           ...
+         </moduleInst>
          <globals>
            ( <globalInst>
                <gAddr>  GADDR </gAddr>
@@ -137,7 +142,12 @@ The operator `#assertLocal`/`#assertGlobal` operators perform a check for a loca
     syntax Auxil ::= "init_global" Int Int
  // --------------------------------------
     rule <k> init_global INDEX GADDR => . ... </k>
-         <globalIndices> GADDRS => GADDRS [ INDEX <- GADDR ] </globalIndices>
+         <curModIdx> CUR </curModIdx>
+         <moduleInst>
+           <modIdx> CUR </modIdx>
+           <globalIndices> GADDRS => GADDRS [ INDEX <- GADDR ] </globalIndices>
+           ...
+         </moduleInst>
          <globals>
            ( .Bag =>
              <globalInst>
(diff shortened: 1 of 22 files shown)


hjorthjort:
For reference, here is the grammar for the reference interpreter:

```
script: <cmd>*

cmd:
  <module>                                   ;; define, validate, and initialize module
  ( register <string> <name>? )              ;; register module for imports
module with given failure string
  <action>                                   ;; perform action and print results
  <assertion>                                ;; assert result of an action
  <meta>                                     ;; meta command

module:
  ...
  ( module <name>? binary <string>* )        ;; module in binary format (may be malformed)
  ( module <name>? quote <string>* )         ;; module quoted in text (may be malformed)

action:
  ( invoke <name>? <string> <expr>* )        ;; invoke function export
  ( get <name>? <string> )                   ;; get global export

assertion:
  ( assert_return <action> <expr>* )         ;; assert action has expected results
  ( assert_return_canonical_nan <action> )   ;; assert action results in NaN in a canonical form
  ( assert_return_arithmetic_nan <action> )  ;; assert action results in NaN with 1 in MSB of fraction field
  ( assert_trap <action> <failure> )         ;; assert action traps with given failure string
  ( assert_malformed <module> <failure> )    ;; assert module cannot be decoded with given failure string
  ( assert_invalid <module> <failure> )      ;; assert module is invalid with given failure string
  ( assert_unlinkable <module> <failure> )   ;; assert module fails to link
  ( assert_trap <module> <failure> )         ;; assert module traps on instantiation

meta:
  ( script <name>? <script> )                ;; name a subscript
  ( input <name>? <string> )                 ;; read script or module from file
  ( output <name>? <string>? )               ;; output module to stout or file
```

test.md
```diff
@@ -258,6 +295,35 @@ We allow to `invoke` a function by its exported name in the test code.
...
+                   |     "register" String Int
+ // ------------------------------------------
+    rule <k> ( register S ) => register S (NEXT -Int 1) ... </k> // Register last module.
+         <nextModuleIdx> NEXT </nextModuleIdx>
+         requires NEXT >Int 0
+    rule <k> ( register S ID:Identifier ) => register S IDX ... </k>
+         <moduleIds> ... ID |-> IDX ... </moduleIds>
+    rule <k> register S:String IDX:Int => . ... </k>
+         <moduleInstances> ... IDX |-> INST ... </moduleInstances>
+         <moduleRegistry> ... .Map => S |-> INST ... </moduleRegistry>
```
QianyangPeng: Do we really need to copy INST? Can we just store the IDX here?

test.md
```diff
@@ -258,6 +295,35 @@ We allow to `invoke` a function by its exported name in the test code.
...
+         <moduleInstances> ... .Map => NEXT |-> INST ... </moduleInstances>
+```
+
+We will reference modules by name in imports.
+`register` is the instruction that allows us to associate a name with a module.
+
+```k
+    syntax Auxil ::= "(" "register" String            ")"
+                   | "(" "register" String Identifier ")"
+                   |     "register" String Int
```
QianyangPeng: Can we just use `"(" "register" String TextFormatIdx ")"` for this rule rather than separately declaring it for `Identifier` and `Int`?

wasm.md
```diff
@@ -1194,12 +1198,24 @@ Module Declaration
...
 
 ```k
     syntax Stmt ::= "(" "module" Defns ")"
- // --------------------------------------
-    rule <k> ( module DEFNS ) => DEFNS ... </k>
+                  | "(" "module" Identifier Defns ")"
+ // -------------------------------------------------
+    rule <k> ( module ID:Identifier DEFNS ) => ( module DEFNS ) ... </k>
+         <moduleIds> ... .Map => ID |-> NEXT ... </moduleIds>
+         <nextModuleIdx> NEXT </nextModuleIdx>
```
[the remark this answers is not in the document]
ehildenb: We should probably be incrementing `NEXT` here if we're using it. So it should be `<nextModuleIdx> NEXT => NEXT +Int 1 </nextModuleIdx>`

ehildenb:
I like @QianyangPeng 's suggestion for storing a reference to the current module and allocating a new empty module as soon as we see a `( module ... )` declaration.

I also like that we can represent the "test-suite embedder" as a separate module, so that `wasm.md` remains the pure Wasm semantics. Note that we don't need to implement all of the semantics that @hjorthjort posted above immediately. We can just give the syntax and add semantics as we start executing more of the test-cases.

test.md
hjorthjort: Absolutely right. Actually, before I pushed the fix for this, the tests were failing.

test.md
hjorthjort: Sure, that makes sense.

test.md
```diff
@@ -258,6 +295,35 @@ We allow to `invoke` a function by its exported name in the test code.
          <exports> ... ENAME |-> TFIDX ... </exports>
 ```
 
+### Registering Modules
+
+We first define how modules get stored away.
+
+```k
+    rule <k> #storeInstance => . ... </k>
+         <moduleInst> INST </moduleInst>
```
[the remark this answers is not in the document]
hjorthjort: I thought `moduleInst` only contained the module being currently defined. If that's the case, this stores the current module, right?

wasm.md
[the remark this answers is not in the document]
hjorthjort: `NEXT` gets incremented when we store the module. This is just reading what the current module index is. Maybe the cell should be called `currentModuleIdx`?

hjorthjort:
Re: storing a reference to the current module, I agree it sounds good. I'll implement it. It will require touching all the top-level rules and change the configuration in large ways, so it might take some wrestling to merge with any current work you have lying around. But on the other hand, it captures the idea of "Frame" from the Wasm spec, so it brings us closer to the spec, which of course always warms my heart.

wasm.md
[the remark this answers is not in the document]
QianyangPeng: I think it should still be next, to keep it consistent with the other `<nXXX>` cells. I think the implementation should be fine, but it is cleaner and more atomic if `<nextModuleIdx>` is used only when it is used.

QianyangPeng:
Yes indeed. We will need to update each `Defn`

hjorthjort:
Re: storing a reference to the current module, I agree it sounds good. I'll implement it. It will require touching all the top-level rules and change the configuration in large ways, so it might take some wrestling to merge with any current work you have lying around. But on the other hand, it captures the idea of "Frame" from the Wasm spec, so it brings us closer to the spec, which of course always warms my heart.

test.md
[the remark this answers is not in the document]
QianyangPeng: Yeah maybe I missed something. Did you empty the `<moduleInst>` somewhere when new module is defined?

hjorthjort:
I think I'm close to have it working. I'll see if I can finish up and, if not, will at least push up what I have so that you can look at it and/or complete it. Since I'll be gone tomorrow, feel free to take over this PR and finish it if you feel you want to work from it.

wasm.md
```diff
@@ -1186,22 +1286,82 @@ Start Function
...
+               <types>         .Map </types>
+               <funcIndices>   .Map </funcIndices>
+               <tabIndices>    .Map </tabIndices>
+               <memIndices>    .Map </memIndices>
+               <globalIndices> .Map </globalIndices>
+               <exports>       .Map </exports>
+             </moduleInst>
+           )
+         </moduleInstances>
+```
```
QianyangPeng: Wow this looks awesome!

test.md
```diff
@@ -20,8 +20,45 @@ This subsort contains Auxiliary functions that only used in our KWASM semantics
...
+We will reference modules by name in imports.
+`register` is the instruction that allows us to associate a name with a module.
+
+```k
+    syntax Auxil ::= "(" "register" String               ")"
+                   | "(" "register" String TextFormatIdx ")"
+ // --------------------------------------------------------
+    rule <k> ( register S ) => ( register S (NEXT -Int 1) )... </k> // Register last instantiated module.
+         <nextModuleIdx> NEXT </nextModuleIdx>
+         requires NEXT >Int 0
```
ehildenb: `r` in `requires` lines up with `l` in `rule`.

Also, let's get rid of the inline comment at the end of line 51.

wasm.md
```diff
@@ -689,12 +706,22 @@ Type could be declared explicitly and could optionally bind with an identifier.
...
-         <nextTypeIdx> NEXTIDX => NEXTIDX +Int 1 </nextTypeIdx>
-         <types>       TYPES   => TYPES [NEXTIDX <- asFuncType(TDECLS)]   </types>
+         <curModIdx> CUR </curModIdx>
+         <moduleInst>
+           <modIdx> CUR </modIdx>
+           <nextTypeIdx> NEXTIDX => NEXTIDX +Int 1 </nextTypeIdx>
+           <types>       TYPES   => TYPES [NEXTIDX <- asFuncType(TDECLS)]   </types>
+           ...
+         </moduleInst>
     rule <k> (type ID:Identifier (func TDECLS:TypeDecls)) => . ... </k>
```
ehildenb: put a blank line before this rule since it's so big

wasm.md
```diff
@@ -689,12 +706,22 @@ Type could be declared explicitly and could optionally bind with an identifier.
                       | "(type" Identifier "(" "func" TypeDecls ")" ")"
  // -------------------------------------------------------------------
     rule <k> (type (func TDECLS:TypeDecls)) => . ... </k>
-         <nextTypeIdx> NEXTIDX => NEXTIDX +Int 1 </nextTypeIdx>
-         <types>       TYPES   => TYPES [NEXTIDX <- asFuncType(TDECLS)]   </types>
+         <curModIdx> CUR </curModIdx>
+         <moduleInst>
+           <modIdx> CUR </modIdx>
+           <nextTypeIdx> NEXTIDX => NEXTIDX +Int 1 </nextTypeIdx>
+           <types>       TYPES   => TYPES [NEXTIDX <- asFuncType(TDECLS)]   </types>
```
ehildenb: spacing here is a bit weird, can we remove the extra spaces before `</types>` or something?

wasm.md
```diff
@@ -927,10 +984,21 @@ Currently, only one memory may be accessible to a module, and thus the `<mAddr>`
...
 
     rule <k> memory { _ _ } => trap ... </k>
-         <memIndices> MAP </memIndices> requires MAP =/=K .Map
+         <curModIdx> CUR </curModIdx>
+         <moduleInst>
+           <modIdx> CUR </modIdx>
+           <memIndices> MAP </memIndices>
+           ...
+         </moduleInst>
+         requires MAP =/=K .Map
```
ehildenb: back up `requires` 2 spaces to line up with `l` in `rule`.

wasm.md
```diff
@@ -1186,22 +1284,82 @@ Start Function
...
+ // Map the given id to the next available module number, and then define the module as usual.
+    rule <k> ( module ID:Identifier DEFNS ) => ( module DEFNS ) ... </k>
+         <moduleIds> ... .Map => ID |-> NEXT ... </moduleIds>
+         <nextModuleIdx> NEXT </nextModuleIdx>
     rule <k> ( module DEFNS ) => DEFNS ... </k>
+         <curModIdx> _ => NEXT </curModIdx>
+         <nextModuleIdx> NEXT => NEXT +Int 1 </nextModuleIdx>
+         <moduleInstances>
+           (.Bag =>
+             <moduleInst>
```
ehildenb: Format as:

```
  ( .Bag
 => <moduleInst>
    ...
  )
```

wasm.md
```diff
@@ -1186,22 +1284,82 @@ Start Function
...
 
+Note that we want to always have a module available, for execution Statements outside of modules.
+We do this with a small hack to ensure that a first module is always created when needed, if not available.
+
+```k
+    rule <curModIdx> 0 </curModIdx>
+         <nextModuleIdx> 0 => 1 </nextModuleIdx>
+         <moduleInstances>
+           (.Bag =>
+             <moduleInst>
```
ehildenb: same formatting here.

Also, I think you can just do:

```
  ( .Bag
 => <moduleInst>
      <modIdx> 0 </modIdx>
      ...
    </moduleInst>
  )
```

and the remaining cells will get default values from the `configuration` declaration, which makes it more robust to changes in the `configuration`.

wasm.md
ehildenb: Actually, as below, we can do:

```
  ( .Bag
 => <moduleInst>
      <modIdx> NEXT </modIxd>
      ...
    </moduleInst>
  )
```

and the remaining cells will get defaults.

PR #105, opened 2019-06-19
Put trap semantics into WASM module now that we can do it properly

hjorthjort:
The fact that `trap` eats up execution is really part of the Wasm semantics. I want to be as principled as possible about the `WASM` module to contain the full Wasm semantics, and to keep embedder stuff located outside. 

Now that we have this nice sort hierarchy, we can finally put `trap` semantics in `WASM`. Do you agree?

### DIFF
--- a/wasm.md
+++ b/wasm.md
@@ -138,7 +138,6 @@ Thus, a `trap` "bubbles up" (more correctly, to "consumes the continuation") unt
     rule <k> trap ~> (IS:Instrs    => .) ... </k>
     rule <k> trap ~> (D:Defn       => .) ... </k>
     rule <k> trap ~> (DS:Defns     => .) ... </k>
-    rule <k> trap ~> ((module  _)  => .) ... </k>
 
     rule <k> trap ~> (S:Stmt SS:Stmts => S ~> SS) ... </k>
 ```
--- a/wasm.md
+++ b/wasm.md
@@ -132,12 +132,12 @@ Thus, a `trap` "bubbles up" (more correctly, to "consumes the continuation") unt
 ```k
     syntax Instr ::= "trap"
  // -----------------------
-    rule <k> trap ~> (L:Label      => .) ... </k>
-    rule <k> trap ~> (F:Frame      => .) ... </k>
-    rule <k> trap ~> (I:Instr      => .) ... </k>
-    rule <k> trap ~> (IS:Instrs    => .) ... </k>
-    rule <k> trap ~> (D:Defn       => .) ... </k>
-    rule <k> trap ~> (DS:Defns     => .) ... </k>
+    rule <k> trap ~> (L:Label   => .) ... </k>
+    rule <k> trap ~> (F:Frame   => .) ... </k>
+    rule <k> trap ~> (I:Instr   => .) ... </k>
+    rule <k> trap ~> (IS:Instrs => .) ... </k>
+    rule <k> trap ~> (D:Defn    => .) ... </k>
+    rule <k> trap ~> (DS:Defns  => .) ... </k>
 
     rule <k> trap ~> (S:Stmt SS:Stmts => S ~> SS) ... </k>
 ```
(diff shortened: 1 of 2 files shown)


wasm.md
```diff
@@ -123,12 +123,30 @@ The sorts `EmptyStmt` and `EmptyStmts` are administrative so that the empty list
...
 ```k
     syntax Instr ::= "trap"
  // -----------------------
+    rule <k> trap ~> (L:Label      => .) ... </k>
+    rule <k> trap ~> (F:Frame      => .) ... </k>
+    rule <k> trap ~> (I:Instr      => .) ... </k>
+    rule <k> trap ~> (IS:Instrs    => .) ... </k>
+    rule <k> trap ~> (D:Defn       => .) ... </k>
+    rule <k> trap ~> (DS:Defns     => .) ... </k>
+    rule <k> trap ~> ((module  _)  => .) ... </k>
```
QianyangPeng: extra space here

wasm.md
```diff
@@ -123,12 +123,30 @@ The sorts `EmptyStmt` and `EmptyStmts` are administrative so that the empty list
...
+
+    rule <k> trap ~> (S:Stmt SS:Stmts => S ~> SS) ... </k>
```
QianyangPeng: actually it is a little twisty to me.
So if a `trap` is followed by `(invoke 0)`, it will be skipped because `(invoke 0)` is an `Instr`. But if a `trap` is followed by `(invoke "abc")`, it will be stopped because `(invoke "abc")` is an `Auxil`. Is it really what we want?

hjorthjort: Isn't that because `invoke 0` is part of Wasm semantics (it's an administrative instruction), whereas `invoke "abc"` is an `Auxil`, outside of Wasm semantics.

If we want, we can add a rule in `TEST` that says that traps eat `Auxil`, too, as long as they are not assertions, but I think that's wrong. If a `trap` is followed by anything but an `#assertTrap`, then execution will halt, right?

The point is: The spec says that traps propagate to the top in Wasm, and can't be caught from within it. These rules make sure that happens. What happens outside of Wasm (once we leave a module instantiation) is up to us.

hjorthjort:
> Maybe (module _ _) needs to be added too, depending on which PR will be merged first.

Well, `(module _ _)` gets desugared into `(module _)`, so I'd avoid the extra rule.

wasm.md
hjorthjort: Where? I don't see it.

wasm.md
QianyangPeng: I was just a little unsure about it. It should be fine!

wasm.md
QianyangPeng: Do you have 2 spaces between `)  =>`?

QianyangPeng:
Will the rule really apply when `(module  _ _)` is not at the top of k cell? I guess `trap ~> ((module  _ _)` may not be desugared..

wasm.md
hjorthjort: Yes, I want to line up the arrows

hjorthjort:
Good point. Actually, I don't think we need to eat the module from the stack, because if we see a module declaration, we have left Wasm and are in the embedder. So that shouldn't be dealt with here.

wasm.md
QianyangPeng: it still lines up if you remove one space from each line, isn't it?
```syntax Instr ::= "trap"
 // -----------------------
    rule <k> trap ~> (L:Label     => .) ... </k>
    rule <k> trap ~> (F:Frame     => .) ... </k>
    rule <k> trap ~> (I:Instr     => .) ... </k>
    rule <k> trap ~> (IS:Instrs   => .) ... </k>
    rule <k> trap ~> (D:Defn      => .) ... </k>
    rule <k> trap ~> (DS:Defns    => .) ... </k>
    rule <k> trap ~> ((module  _) => .) ... </k>

    rule <k> trap ~> (S:Stmt SS:Stmts => S ~> SS) ... </k>```

hjorthjort: Ah, not I get what you mean. Thanks.

hjorthjort: Fixed.

PR #107, opened 2019-06-21
Cleanup the assertions

QianyangPeng:
now the assertions do not delete elements.
fixed the test file `start.wast`

### DIFF
--- a/test.md
+++ b/test.md
@@ -91,15 +91,20 @@ These functions make assertions about the state of the `<valstack>` cell.
     syntax Assertion ::= "#assertTopStack"        Val      String
                        | "#assertTopStackExactly" Val      String
                        | "#assertStack"           ValStack String
- // -------------------------------------------------------------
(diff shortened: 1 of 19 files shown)


hjorthjort:
A lot cleaner. Thank you!

test.md
```diff
@@ -267,30 +260,22 @@ This asserts related operation about tables.
...
-               <mdata>   .Map  </mdata>
-               ...
-             </memInst>
-          => .Bag
-           )
+           <memInst>
+             <mAddr>   ADDR  </mAddr>
+             <mmax>    MAX   </mmax>
+             <msize>   SIZE  </msize>
+             ...
```
hjorthjort: Small regression here: It should still check that `<mdata>` contains exactly `.Map` (i.e., is empty). I had some bugs where memory bytes spilled over, so the idea with this test was to assert that the memory was, in fact, empty (or at least didn't contain any more bytes than the intended ones. So it would be good to have that check.

PR #113, opened 2019-06-26
(Re)implement proper inline modules, and implement global declarations

hjorthjort:
The spec explicitly allows inline modules, like we had earlier, where definitions could be placed at the top level, i.e., no need to explicitly use `module` keyword. See the official test `inline_module.wast` for an example.

The rule that allows this is the very last one in `wasm.md`, which kicks in only when a `Defn` is encountered and there is no current module. In that case, a module is instantiated.

I'm not a huge fan of the fact that we can (and do) use a simple `(module)` statement to create a module and then have it hanging around as the current module. I'd prefer if we actually were a bit stricter so that the `curModIdx` cell is cleared after a module declaration is done. This messes a lot with our assertions, though, and I haven't really got that working yet.

The new rule relies on grouping `Defns`, so I had some issues with `init_globals` being an `Auxil`. I figured it would be straightforward to just replace it with actual `global` definitions.

### DIFF
--- a/tests/success-haskell.out
+++ b/tests/success-haskell.out
@@ -12,45 +12,25 @@
     <locals>
       .Map
     </locals>
+    <curModIdx>
+      .Map
+    </curModIdx>
   </curFrame>
   <nextFreshId>
     0
   </nextFreshId>
-  <moduleInst>
-    <typeIds>
-      .Map
-    </typeIds>
-    <funcIds>
-      .Map
-    </funcIds>
-    <nextTypeIdx>
-      0
-    </nextTypeIdx>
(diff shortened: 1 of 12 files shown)


tests/success-haskell.out
```diff
@@ -12,45 +12,25 @@
...
-      .Map
-    </memIndices>
-    <globalIndices>
-      .Map
-    </globalIndices>
-    <exports>
-      .Map
-    </exports>
-  </moduleInst>
+  <moduleIds>
```
hjorthjort: This golden file was fairly out-of sync

PR #114, opened 2019-06-27
Structuring and ordering definitions in modules

hjorthjort:
This traverses the definitions in a module and orders them appropriately for allocation and instantiation. Imports are yet to be implemented, but I added them to the ordering for completeness.

### DIFF
--- a/wasm.md
+++ b/wasm.md
@@ -1476,8 +1476,7 @@ The groups are chosen to represent different stages of allocation and instantiat
 
     // Imports (TODO).
  // rule #structureModule(M, (I:ImportDefn DS:Defns))
- //
-      => #structureModule(M ["imports"   <- (I {M ["imports"  ]}:>Defns)], DS)
+ //   => #structureModule(M ["imports"   <- (I {M ["imports"  ]}:>Defns)], DS)
 
     // Allocations.
     rule #structureModule(M, (A:FuncDefn   DS:Defns))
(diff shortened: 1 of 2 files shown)


wasm.md
```diff
@@ -1438,20 +1440,102 @@ Start Function
 Module Instantiation
 --------------------
 
+There is some dependencies among definitions that require that we do them in a certain order, even though they may appear in many valid orders.
+First, functions, tables, memories and globals get *allocated*.
+Then, tables, memories and globals get *instantiated* with elements, data and initialization vectors.
+However, since (currently) globals can only make use of imported globals to be instantiated, we can initialize at allocation time.
+Finally, the start function is evoked.
```
ehildenb: evoked -> invoked?

wasm.md
```diff
@@ -1438,20 +1440,102 @@ Start Function
...
+Exports may appear anywhere in a function, but can only be performed after what they refer to is allocated.
+Inlined imports and exports are safe to extract as they appear.
+Imports must appear first in a module due to validation.
+Exports only depend on what they are exporting having been allocated, so can appear right after allocation.
```
ehildenb: `Exports may appear anywhere in a function ...`, is this talking about `Inlined` exports? If so, it seems to contradict what the next sentence says. If not, maybe clarify what an `Inlined` export/import is.

wasm.md
```diff
@@ -1438,20 +1440,102 @@ Start Function
...
+
+    // Initializations.
+    rule #structureModule(M, (I:DataDefn   DS:Defns))
+      => #structureModule(M ["inits"     <- (I {M ["inits"    ]}:>Defns)], DS)
+    rule #structureModule(M, (I:ElemDefn   DS:Defns))
+      => #structureModule(M ["inits"     <- (I {M ["inits"    ]}:>Defns)], DS)
+
+    // Start function.
+    rule #structureModule(M, (S:StartDefn  DS:Defns))
+      => #structureModule(M ["start"     <- (S .Defns)], DS) // There can only be one start function.
```
ehildenb: I like this, but let's do one rule per line, so that everything lines up nicer. Also we don't need the inline comments (unless they are `TODO`). The sorts here are enough to tell us what each line does, and if there is something which needs to be there for documentation, do it in the markdown above the code block.

wasm.md
hjorthjort: Sorry, that should have been "module", not "function". Does that make it clearer?

wasm.md
hjorthjort: Done

PR #121, opened 2019-07-02
Importing and exporting

hjorthjort:


test.md
```diff
@@ -60,6 +60,22 @@ We allow 2 kinds of actions:
            <exports> ... ENAME |-> TFIDX ... </exports>
            ...
          </moduleInst>
+    rule <k> ( get NAME:String ) => ( get CUR NAME) ... </k>
+         <curModIdx> CUR </curModIdx>
```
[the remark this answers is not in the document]
hjorthjort: Good catch.

tests/simple/imports.wast
```diff
@@ -0,0 +1,27 @@
...
+ (func (import "m" "f"))
+ (global $x (import "m" "g") (mut i32))
+ (func (export "foo") (result i32) (global.get 0))
+ (func (export "mod") (global.set 0 (i32.const 10)))
+ )
+
+(assert_return (invoke "foo") (i32.const 42))
+(invoke "mod")
+(assert_return (invoke "foo") (i32.const 10))
+(assert_return (get $a "g") (i32.const 10))
```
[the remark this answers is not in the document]
hjorthjort: Adding a more full implementation of both `invoke` and `get`

hjorthjort:
This PR merges both #127  and #128, and should wait for those to merge. It is otherwise ready for review again.

tests/simple/imports.wast
```diff
@@ -0,0 +1,27 @@
...
+   )
+
+(register "m")
+
+(module
+ (memory (import "m" "mem") 1)
+ (export "x" (global $x))
+ (type $t (func (param i64) (param i32) (result i32)))
+ (func (import "m" "gunc") (type $t))
+ (func (import "m" "f"))
```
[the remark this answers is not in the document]
hjorthjort: Added `linking.wast` to tests.

wasm.md
```diff
@@ -849,30 +848,36 @@ It could also be declared implicitly when a `TypeUse` is a `TypeDecls`, in this
 
 Function declarations can look quite different depending on which fields are ommitted and what the context is.
 Here, we allow for an "abstract" function declaration using syntax `func_::___`, and a more concrete one which allows arbitrary order of declaration of parameters, locals, and results.
+Functions can either be specified by giving a type, what locals it allocates, and a function body; or by an import and it's expected type.
```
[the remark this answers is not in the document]
hjorthjort: Not sure where to put such an explanation, so I put it under the section `Inline Exports`. We could also deal with all the export desugaring under that section. Let me see how that looks like.

wasm.md
```diff
@@ -1017,23 +1005,43 @@ The only allowed `TableElemType` is "funcref", so we ignore this term in the red
...
+         <nextFreshId> NEXTID => NEXTID +Int 1 </nextFreshId>
+      requires EXPO =/=K .InlineExports
+
+    rule <k> ( table ID:Identifier ( export ENAME ) EXPO:InlineExports SPEC:TableSpec )
+          => ( export ENAME ( table ID ) )
+          ~> ( table ID EXPO SPEC )
+          ...
+         </k>
+
+    rule <k> ( table .InlineExports funcref ( elem ES ) ) => ( table #freshId(NEXTID) .InlineExports funcref (elem ES) ) ... </k>
```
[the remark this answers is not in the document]
hjorthjort: No, it is dealing with the case when we need to desugar the inline `elem` definition, and we don't already have an identifier. So we need a fresh identifier, because the `elem` needs to know which table to refer to (technically, it's not necessary, but the spec puts it this way).

hjorthjort:
I'll also experiment with putting all logic for importing and exporting under their respective sections, to see if that becomes cleaner.

wasm.md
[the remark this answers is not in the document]
hjorthjort: Okay, so the problem is that since an empty list of exports is called `.InlineExports`, it needs to be explicitly mentioned in the rules. So if we move this exporting logic to another section, we have this wierd `.InlineExports` thing hanging around.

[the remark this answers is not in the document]
hjorthjort: I think I have a solution though that will turn out quite nice.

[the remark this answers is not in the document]
hjorthjort: Re: this: now all the exports and imports syntax and rules for desugaring are under their respective sections, so the main part of the semantics actually doesn't deal with them at all. It also becomes more readable I think, because the desugaring rules are almost identical for the different allocated kinds (i.e., functions, globals, etc.)

hjorthjort:
I did a bit of an overhaul and moved all the export and import rules to their respective regions, which cleans up the rest of the semantics a bit. This still is waiting for #128, though. Otherwise ready for review (finally).

wasm.md
```diff
@@ -212,6 +212,7 @@ To resolve these `identifiers` into concrete `indices`, some grammar production
  // ----------------------------------------------------------------
     rule #ContextLookup(IDS:Map, I:Int) => I
     rule #ContextLookup(IDS:Map, ID:Identifier) => {IDS [ ID ]}:>Int
+      requires ID in_keys(IDS)
```
hjorthjort: This came up a lot during testing: the function would succeed, with result `#Bottom`, get cast to `Int` and the error message was not helpful. With this change, at least execution halts when for some reason this doesn't hold (which it should if you're calling this function).

wasm.md
```diff
@@ -1531,6 +1496,238 @@ Start Function
...
+The value of a global gets copied when it is imported.
+
+```k
+    syntax Defn       ::= ImportDefn
+    syntax ImportDefn ::= "(" "import" String String ImportDesc ")"
+    syntax ImportDesc ::= "(" "func"   OptionalId TypeUse        ")" [klabel(funcImportDesc)]
+                        | "(" "table"  OptionalId TableType      ")" [klabel( tabImportDesc)]
+                        | "(" "memory" OptionalId MemType        ")" [klabel( memImportDesc)]
+                        | "(" "global" OptionalId TextGlobalType ")" [klabel(globImportDesc)]
+ // -------------------------------------------------------------------------------------
```
QianyangPeng: syntax here (----- line up with ])

wasm.md
```diff
@@ -1531,6 +1496,238 @@ Start Function
...
+
+Exports can also be declared like regular functions, memories, etc., by giving an inline export declaration.
+In that case, it simply desugars to an export followed by the definition, after introducing a fresh identifier if no identifier is present.
+Note that it is possible to define multiple exports inline, i.e., export a single entity under many names.
+
+```k
+    syntax InlineExport  ::= "(" "export" String ")"
+ // ------------------------------------------------
+
+    syntax GlobalSpec ::= InlineExport GlobalSpec
```
QianyangPeng: I am not a fan of using those `XXXspec` to integrate the normal form of definitions and the abbreviated form of definitions. It makes the coupling between definitions pretty heavy and makes the logic complicated. I prefer to have:
```
syntax InlineDecls ::= InlineExport InlineDecls
syntax GlobalDefn ::= "(" "global" OptionalId InlineExport InlineDecls GlobalSpec ")"
// ---
rule <k> ( global                  EXPO:InlineExport InlineDecls SPEC:GlobalSpec )
          => ( global #freshId(NEXTID) EXPO         InlineDecls      SPEC            )
          ...
         </k>
         <nextFreshId> NEXTID => NEXTID +Int 1 </nextFreshId>
```
But this is just my personal opinion. : )

wasm.md
```diff
@@ -1096,16 +1058,19 @@ Currently there is only one possiblt value for it which is "funcref".
 The allocation of a new `tableinst`.
 Currently at most one table may be defined or imported in a single module.
 The only allowed `TableElemType` is "funcref", so we ignore this term in the reducted sort.
+The table values are addresses into the store of functions.
 
 ```k
     syntax Defn      ::= TableDefn
-    syntax TableDefn ::= "(" "table"     OptionalId Limits TableElemType ")"
-                       | "(" "table"     OptionalId        TableElemType "(" "elem" ElemSegment ")" ")"
+    syntax TableType ::= Limits TableElemType
```
QianyangPeng: I prefer to put the definition of `TableType`, `TypeUse`, `TableType` and `TextGlobalType` together because they are not only used in their own definition, but are also used elsewhere in the program. Putting them together in another section makes people know that "if you change this, you will not only influence this section, but you will also influence somewhere else"

wasm.md
QianyangPeng: Also isn't it weird that we have `TableType`, `MemType`, but `TextGlobalType`? Can we either make them all with `Text` or all without `Text`?

wasm.md
```diff
@@ -1531,6 +1496,238 @@ Start Function
...
+
+```k
+    syntax Defn       ::= ImportDefn
+    syntax ImportDefn ::= "(" "import" String String ImportDesc ")"
+    syntax ImportDesc ::= "(" "func"   OptionalId TypeUse        ")" [klabel(funcImportDesc)]
+                        | "(" "table"  OptionalId TableType      ")" [klabel( tabImportDesc)]
+                        | "(" "memory" OptionalId MemType        ")" [klabel( memImportDesc)]
+                        | "(" "global" OptionalId TextGlobalType ")" [klabel(globImportDesc)]
+ // -------------------------------------------------------------------------------------
+    rule <k> ( import MOD NAME (func OID:OptionalId TUSE:TypeUse)) => . ... </k>
```
QianyangPeng: you have space after first paren but not before the last paren.

wasm.md
```diff
@@ -1531,6 +1496,238 @@ Start Function
...
+           ...
+         </moduleInst>
+         <funcDef>
+           <fAddr> ADDR </fAddr>
+           <fType> FTYPE </fType>
+           ...
+         </funcDef>
+      requires unnameFuncType(FTYPE) ==K unnameFuncType(asFuncType(TYPEIDS, TYPES, TUSE))
+
+    rule <k> ( import MOD NAME (table OID:OptionalId (LIM _):TableType)) => . ... </k>
```
QianyangPeng: same. you have space after first paren but not before the last paren.

wasm.md
```diff
@@ -1531,6 +1496,238 @@ Start Function
...
+         </moduleInst>
+         <tabInst>
+           <tAddr> ADDR </tAddr>
+           <tmax>  MAX  </tmax>
+           <tsize> SIZE </tsize>
+           ...
+         </tabInst>
+       requires #limitsMatchImport(SIZE, MAX, LIM)
+
+    rule <k> ( import MOD NAME (memory OID:OptionalId (LIM:Limits):MemType)) => . ... </k>
```
QianyangPeng: same

wasm.md
```diff
@@ -1531,6 +1496,238 @@ Start Function
...
+         </moduleInst>
+         <memInst>
+           <mAddr> ADDR </mAddr>
+           <mmax>  MAX  </mmax>
+           <msize> SIZE </msize>
+           ...
+         </memInst>
+       requires #limitsMatchImport(SIZE, MAX, LIM)
+
+    rule <k> ( import MOD NAME (global OID:OptionalId TGTYP:TextGlobalType)) => . ... </k>
```
QianyangPeng: same

wasm.md
```diff
@@ -1531,6 +1496,238 @@ Start Function
...
+ // -------------------------------------------------
+    rule <k> ( global OID:OptionalId (import MOD NAME) TYP ) => ( import MOD NAME (global OID TYP) ) ... </k>
+
+    syntax FuncSpec ::= InlineImport TypeUse
+ // ----------------------------------------
+    rule <k> ( func OID:OptionalId (import MOD NAME) TUSE ) => ( import MOD NAME (func OID TUSE) ) ... </k>
+
+    syntax TableSpec ::= InlineImport TableType
+ // -------------------------------------------
+    rule <k> ( table OID:OptionalId (import MOD NAME) TT:TableType ) => (import MOD NAME (table OID TT)) ... </k>
```
QianyangPeng: syntax of `(import MOD NAME (table OID TT))` is different with `( import MOD NAME (func OID TUSE) )` (whether to have spaces)

wasm.md
```diff
@@ -1531,6 +1496,238 @@ Start Function
...
+
+    syntax MemorySpec ::= InlineImport MemType
+ // ------------------------------------------
+    rule <k> ( memory OID:OptionalId (import MOD NAME) MT:MemType ) => (import MOD NAME (memory OID MT)) ... </k>
```
QianyangPeng: same here.

wasm.md
```diff
@@ -1531,6 +1496,238 @@ Start Function
          </moduleInst>
 ```
 
+Export
+------
+
+Exports make functions, tables, memories and globals available for importing into other modules.
```
QianyangPeng: hmmm.. it also makes it available to be accessed by `Actions`. You decide whether to mention because `Action` belongs to `test.md`.

hjorthjort: I prefer to not mention actions here, like you say, because it belongs to `test.md`, i.e. the embedder.

wasm.md
hjorthjort: Yeah, I don't think it's super elegant either, but it's the best I could come up with.

The problem is that this is one of the places where abstract and concrete syntax diverge. `TextGlobalType` is concrete, and `GlobalType` is abstract. The abstract syntax annotates every global with `var` or `const`, but the concrete just puts `mut` on the mutable and nothing on the constant values. So if we want to use `var` and `const` in our rules (which I think we should), we need to differentiate them from the text syntax (which we need to parse) which just uses `mut`. For the other types the abstract and constant syntaxes are the same, so naming them `Text*` would be strange (since it's not specifically concrete syntax).

wasm.md
hjorthjort: Now that you mention it, the exports aren't really part of the "specification" of globals/functions/tables/mems, so yeah, that's a bit weird.

What would `InlineDecls` be in this case? What makes it different? Would it be like what we had before, with a list type `InlineExports`, only that it's non-empty? So it could also be something like this:

```
    syntax InlineExport   ::= "(" "export" String ")"
    syntax InlineExports  ::= InlineExport | InlineExport InlineExports

    syntax GlobalDefn ::= "(" "global" OptionalId InlineExports GlobalSpec ")"
 // ---------------------------------------------
 ...
```

The problem is that we get more rules: one to deal with the base case (singleton list), one to deal with the inductive case, and one to introduce identifiers. So in total, we get more rules: one for functions, one for tables, etc.

My thinking here was that using `syntax GlobalSpec ::= InlineExport GlobalSpec` we get kinda close to the spec, where they give these inline exports (see page 118 for example) as `‘(’ ‘func’ id? ‘(’ ‘export’ name ‘)’ ... ‘)’` and the explanation "The latter abbreviation can be applied repeatedly, with “. . .” containing another import or export."

Maybe there's a better naming than `*Spec`?

wasm.md
hjorthjort: I'm not sure I follow. What's the connection between `TypeUse` on the one hand and `TextGlobalType` and `TableType` on the other. Do you mean a section about all the "types" `GlobalType`, `MemType`, `FuncType` and `TableType`?

wasm.md
hjorthjort: Fixed.

wasm.md
hjorthjort: Fixed

wasm.md
hjorthjort: Fixed

wasm.md
hjorthjort: Fixed

wasm.md
hjorthjort: Fixed

wasm.md
hjorthjort: Fixed

wasm.md
```diff
@@ -689,9 +688,10 @@ When globals are declared, they must also be given a constant initialization val
     rule asGMut (      T:AValType   ) => const T
 
     syntax Defn       ::= GlobalDefn
-    syntax GlobalDefn ::= "(" "global" OptionalId TextGlobalType Instr ")"
-                        |     "global" GlobalType
- // ---------------------------------------------
+    syntax GlobalSpec ::= TextGlobalType Instr
+    syntax GlobalDefn ::= "(" "global" OptionalId GlobalSpec ")"
+                        |     "global" OptionalId GlobalType
+ // --------------------------------------------------------
```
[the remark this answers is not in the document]
hjorthjort: I agree it would be good to document this, but I'm not sure where. These are not just local idioms: the `*Spec` production is used for all the allocated entities (func/mem/tab/glob) because they have similar structure of definition, `Text*` as an indicator that something really just belongs to the text format is in use with `TextFormatIdx`, and we have lots of abstract helpers in many places. I'll put some explanation here, but make a note to reconsider it when we do a more complete overview of the semantics.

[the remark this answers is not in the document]
hjorthjort: I added some documentation here, and similar documentation to the other sections where a `*Spec` is introduced. What do you think?

Also, would you prefer if `TextGlobalType` was called `TextFormatGlobalType`?

wasm.md
QianyangPeng: Yes I just mean having a separated section for all the `types`

wasm.md
QianyangPeng: How about:
TextGlobalType -> GlobalType
GlobalType -> AbstractGlobalType?
This looks a little easier to understand (and more consistent) to me.

QianyangPeng: Or you can use `TextFormatGlobalType`, this looks a little better than `TextGlobalType`

wasm.md
QianyangPeng: Yes this will introduce more rules. Maybe using `GlobalSpec` as you did is fine. Better to get confirmation from @ehildenb

test.md
```diff
@@ -47,19 +47,48 @@ We allow 2 kinds of actions:
...
-    rule <k> ( invoke ENAME:String )           => ( call TFIDX )         ... </k>
+    syntax Action ::= "(" "invoke" OptionalId String Instrs ")"
+                    |     "invoke" Int        String
+                    | "(" "get"    OptionalId String        ")"
+                    |     "get"    Int        String
+ // ------------------------------------------------
+    rule <k> ( invoke OID:OptionalId ENAME:String IS:Instrs ) => IS ~> ( invoke OID ENAME .Instrs ) ... </k>
+      requires IS =/=K .Instrs
+
+    rule <k> ( invoke ENAME:String .Instrs ) => invoke CUR ENAME ... </k>
```
ehildenb: Let's use the actual klabel, lo LHS looks like `( invoke .Identifier ENAME:String .Instrs )`

wasm.md
```diff
@@ -1531,6 +1510,238 @@ Start Function
...
+         </moduleInst>
+         <tabInst>
+           <tAddr> ADDR </tAddr>
+           <tmax>  MAX  </tmax>
+           <tsize> SIZE </tsize>
+           ...
+         </tabInst>
+       requires #limitsMatchImport(SIZE, MAX, LIM)
+
+    rule <k> ( import MOD NAME (memory OID:OptionalId (LIM:Limits):MemType) ) => . ... </k>
```
ehildenb: What's up with `(LIM:Limits):MemType)`? AFAICT, `Limits` is the only subsort of `MemType`, so why do we have the intermediate sort `MemType`. But why are we sort checking it twice?

wasm.md
```diff
@@ -1531,6 +1510,238 @@ Start Function
...
+           <globalIndices> GS => GS [NEXT <- ADDR] </globalIndices>
+           <nextGlobIdx> NEXT => NEXT +Int 1 </nextGlobIdx>
+           ...
+         </moduleInst>
+         <moduleRegistry> ... MOD |-> MODIDX ... </moduleRegistry>
+         <moduleInst>
+           <modIdx> MODIDX </modIdx>
+           <globIds> IDS' </globIds>
+           <globalIndices> ... #ContextLookup(IDS' , TFIDX) |-> ADDR ... </globalIndices>
+           <exports>     ... NAME |-> TFIDX                          ... </exports>
```
ehildenb: formatting `...`.

test.md
```diff
@@ -47,19 +47,48 @@ We allow 2 kinds of actions:
...
          <moduleInst>
-           <modIdx> CUR </modIdx>
+           <modIdx> MODIDX </modIdx>
            <exports> ... ENAME |-> TFIDX ... </exports>
+           <funcIds> IDS </funcIds>
+           <funcIndices> ... #ContextLookup(IDS , TFIDX) |-> FADDR ... </funcIndices>
            ...
          </moduleInst>
+
+    rule <k> ( get NAME:String ) => get CUR NAME  ... </k>
```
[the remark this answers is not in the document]
ehildenb: We found that this doesn't work.

test.md
ehildenb: We found that this doesn't work.

wasm.md
hjorthjort: Yeah, that's a mistake. Fixed.

wasm.md
hjorthjort: Fixed.

wasm.md
hjorthjort: Okay, `TextFormatGlobalType` makes sense to me.

Also, I just realized that in the spec, all these productions have the same name, e.g., there are two productions called `globaltype`, one for the text format and one for the abstract syntax, and the only difference is the font. So we should maybe adopt some convention for differentiating them. I think `TextFormat*` becomes a bit long if we use it in many places, so maybe, `TF*` as a prefix where necessary would be better?

wasm.md
hjorthjort: Good, I'll make a note of that for refactoring. Probably best to avoid doing big file rearrangements in this PR.

PR #127, opened 2019-07-04
Bugfix: `elem` should initialize with function addresses, not ids

hjorthjort:
Previusly we were storing identifiers, but two modules may use the same
identifier for the same function, and the lookup would thus break.

For example
```
(module $0
  (table (export "tab") 5 funcref)
  (type (func (result i32)))
  (func $f)
  (func (export "call") (type 0) (call_indirect (i32.const 0)))
)

(register $0 "Mt")

(module $1
  (table (import "Mt" "tab") 5 funcref)
  (elem (i32.const 0) $f)
  (func $f (result i32) (i32.const 42))
)

(invoke $0 "call)
```

This should put the function `$f` in from *current* module in the
table. Note, though, the module where the table is stored may call into the
table. If the table stored identifiers, the `call_indirect` instruction would find the id `$f` and call the function of that name defined in $0, which is incorrect.


QianyangPeng:
Just a small question. How do you know that the example above will fail? I agree it looks weird but at least. the reference interpreter doesn't accept such a program at all, even after I do some syntax fixing.
```
(module $0
  (table (export "tab") 5 funcref)
  (type (func (result i32)))
  (func $f)
  (func (export "call") (type 0) (call_indirect (i32.const 0)))
)

(register "Mt" $0)

(module $1
  (table (import "Mt" "tab") 5 funcref)
  (elem (i32.const 0) $f)
  (func $f (result i32) (i32.const 42))
)

(invoke "call" $0)
```

Can you provide an example that shows this issue which is executable with the reference interpreter?

hjorthjort:
> (module $0
>   (table (export "tab") 5 funcref)
>   (type (func (result i32)))
>   (func $f)
>   (func (export "call") (type 0) (call_indirect (i32.const 0)))
> )
> 
> (register "Mt" $0)
> 
> (module $1
>   (table (import "Mt" "tab") 5 funcref)
>   (elem (i32.const 0) $f)
>   (func $f (result i32) (i32.const 42))
> )
> 
> (invoke "call" $0)

Sorry, here's a working version (forgot to give the type to `call_indirect`)

```
(module $0
  (table (export "tab") 5 funcref)
  (type (func (result i32)))
  (func $f)
  (func (export "call") (type 0) (call_indirect (type 0) (i32.const 0)))
)

(register "Mt" $0)

(module $1
  (table (import "Mt" "tab") 5 funcref)
  (elem (i32.const 0) $f)
  (func $f (result i32) (i32.const 42))
)

(invoke $0 "call")
```

wasm.md
```diff
@@ -1447,28 +1443,30 @@ A table index is optional and will be default to zero.
...
+           <funcIndices> FADDRS </funcIndices>
+           <tabIds>  TIDS </tabIds>
+           <tabIndices> #ContextLookup(TIDS, TABIDX) |-> ADDR </tabIndices>
            ...
          </moduleInst>
 
-    rule <k> #initElements ( ADDR, OFFSET, .ElemSegment ) => . ... </k>
-    rule <k> #initElements ( ADDR, OFFSET,  E ES        ) => #initElements ( ADDR, OFFSET +Int 1, ES ) ... </k>
+    rule <k> #initElements (    _,     _ ,      _,   _, .ElemSegment ) => . ... </k>
+    rule <k> #initElements ( ADDR, OFFSET, FADDRS, IDS,  E ES        ) => #initElements ( ADDR, OFFSET +Int 1, FADDRS, IDS, ES ) ... </k>
```
QianyangPeng: Not sure as this function is in `<k>` cell, why do we need to pass `FADDRS` and `IDS` to the function, rather than directly look it up in the configuration?

hjorthjort: I figured that if we are using a helper `Stmt`, we might as well give it as little context as possible. (Basically, it just makes the rule smaller, as it is now acting just on the store and not the module it is in).

PR #130, opened 2019-07-05
"Correct" filename for kompiling to Haskell

hjorthjort:
Kompiling to haskell no longer produces a `kore.txt` file, but rather a `compiled.txt` file. Due to the nature of Makefiles, this issue meant that make constantly found the `kore.txt` file lacking, and thus always rebuilt the haskell backend regardless of if it's necessary or not. That's a lot of wasted time every time `make build` is called.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,7 +58,7 @@ java_kompiled:=$(java_dir)/test-kompiled/compiled.txt
 
 haskell_dir:=$(defn_dir)/haskell
 haskell_defn:=$(patsubst %, $(haskell_dir)/%, $(wasm_files))
-haskell_kompiled:=$(haskell_dir)/test-kompiled/compiled.txt
+haskell_kompiled:=$(haskell_dir)/test-kompiled/definition.kore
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -58,7 +58,7 @@ java_kompiled:=$(java_dir)/test-kompiled/compiled.txt
 
 haskell_dir:=$(defn_dir)/haskell
 haskell_defn:=$(patsubst %, $(haskell_dir)/%, $(wasm_files))
-haskell_kompiled:=$(haskell_dir)/test-kompiled/kore.txt
+haskell_kompiled:=$(haskell_dir)/test-kompiled/compiled.txt
```
ehildenb: The haskell backend does use `kore.txt` as the canonical source, not `compiled.txt`. Actually any of the backends using `compiled.txt` as the target file is probably not the best idea, since it's not actually used for anything.

ehildenb: But there isn't much of a better choice for the Java backend.

PR #137, opened 2019-07-10
Cleanups

hjorthjort:
* Remove trailing whitespace
* Fix alignment
* Clean up a rule for globals
* Remove unnecessary side condition
* Add side condition to `#ContextLookup` so that execution halts rather than blow up with weird error messages.


wasm.md
```diff
@@ -819,7 +814,7 @@ A type use should start with `'(' 'type' x:typeidx ')'` followed by a group of i
 ### Type Declaration
 
 Type could be declared explicitly and could optionally bind with an identifier.
-The `identifier` for `param` of the keyword will be used 
+The `identifier` for `param` of the keyword will be used.
```
hjorthjort: @QianyangPeng I think this sentence was never properly completed. Feel free to fix it.

QianyangPeng: ah yes. It is an outdated sentence. Now the `identifier` for `param` of the keyword will be discarded because it is not useful. I have updated it.

PR #138, opened 2019-07-10
Configuration changes

hjorthjort:
NOTE: This PR is based on #137, so should wait for it.

This restructures the configuration in a somewhat principled manner. 
* Rename `funcDef` to `funcInst`, to match the other instances.
* Make configuration read easily top to bottom when performing a call/invoke/import across modules:
  * The module registry is where you look to find the module name you are calling into
  * The registry points to a module index
  * The module begins with listing its exports, which is what you are calling/making use of
  * The export points to an identifier
  * The identifier points to an address
  * The addresses are found further down, in the store.
* Put all "helper" cells for `next*Idx` and so on after the other cells in its group. The reasoning is that they are mostly for bookkeeping, and not that interesting when scanning to try to understand what is going on/should be going on.

I found this structure to be very helpful when debugging imports and exports. As programs grow large, scrolling up and down a lot to follow indirections becomes a nuisance. Being able to at least mostly flow downwards to find what you are calling/importing/invoking at least makes life a little easier.


wasm.md
```diff
@@ -204,14 +204,15 @@ Function `#unsigned` is called on integers to allow programs to use negative num
 
 ### Text Format Conventions
 
-The text format allows the use of symbolic `identifiers` in place of `indices`.
-To resolve these `identifiers` into concrete `indices`, some grammar production are indexed by an identifier context `I` as a synthesized attribute that records the declared identifiers in each index space. We call this operation `ICov`.
+The text format allows the use of symbolic `identifiers` in place of `addrs`.
```
QianyangPeng: Hmm.. What does it mean? I think the symbolic `identifiers` are representing the `index` in the module instance. `addr` should be in the store and has nothing to do with the `identifier`

hjorthjort: Ah, good catch. This seems like the result of a global search&replace.

hjorthjort: It seemed the documentation was slightly out of date here, so I updated it and extended it a bit.

Issue #139, opened 2019-07-10
TODO list of `Float`


QianyangPeng:
Here is a list of float related functionalities that are tricky to implement with our the current design.

Parsing
--------
1. Negative NaN (-nan): This should represents a `canonical NaN` whose sign bit is 1.
2. Arithmetic NaN (&plusmn;nan:0x): When "&plusmn;nan" is followed by a colon `:`, then it shall be followed by a `hexnum` representing the `payload` of corresponding NaN.
3. Hexfloat (‘0x’ 𝑝:hexnum ‘.’ 𝑞:hexfrac (‘P’ | ‘p’) ±:sign 𝑒:num):  This is a Float represented in hexadecimal form.

Currently in `K` I don't see an approach to control the `sign bit` or the `payload` of an `NaN`, so maybe there is no direct way to implement negative NaN and Arithmetic NaN.

Conversions
------------
1. Reinterpret (i32.reinterpret_f32, i64.reinterpret_f64, f32.reinterpret_i32, f64.reinterpret_i64): These operations cast the type of a `val` between int and float but keep the binary representation. For example, `i32.reinterpret_f32 (f32.const 1.5) = <i32> 1069547520`

Memory
--------
1. Store (fxx.store): This should store a float number into memory byte by byte.
2. Load (fxx.load): This should load a float number from memory.

If you have any idea how to implement any of the above, please comment below, thanks!

dwightguth:
mpfr doesn't support NaN payloads, so we would have to implement this by means of using system floats and doubles. It's doable, but requires significant work on the K backend, and would only work for these two bit widths of floats.

My advice would be to implement this functionality in the llvm and haskell backends and try to avoid implementing it in the java or ocaml backends.

hjorthjort:
What does the spec say about nondeterminism when dealing with NaN? If we don't have to be able to perform deterministic computations on them we could just represent them as bytes (`Int`s), which also makes storing to memory and reinterpreting simple.

QianyangPeng:
@hjorthjort Can you elaborate how to represent them as bytes (Ints)? I think we have to perform deterministic computations on them though.

hjorthjort:
> @hjorthjort Can you elaborate how to represent them as bytes (Ints)? I think we have to perform deterministic computations on them though.

[...]

PR #144, opened 2019-07-11
Float impl

QianyangPeng:
1. Modified the makefile and the test cases under tests/simple/float will be executed with java backend.
2. assert_return changed from #assertTopStackExactly back to #assertTopStack, and implemented it correctly.
3. Add a script convert.py. We can use it to convert the hexfloats in a conformance test to a normal float, and it can comment out the unimplemented lines.
4. Implemented all float related operations, except for the ones specified here: https://github.com/kframework/wasm-semantics/issues/139

**when running `tests/simple/float/fconversions-c.wast` it has a warning `[Warning] Internal: Ignored exception thrown by hook Float2Int (Rounding necessary)`.

It's because *the java backend doesn't short circuit evaluate boolean operations*. Rather than fixing the warning on the semantics side, I'd prefer there will be a backend fix in the future.


tests/simple/conversion.wast
```diff
@@ -24,4 +24,24 @@
 (i64.extend_i32_s (i32.const 15))
 #assertTopStack < i64 > 15 "folded extend sig"
 
+(module
```
hjorthjort: Is this a conformance test? Maybe better to just give it its own file, then?

tests/simple/float/f64-cs.wast
```diff
@@ -0,0 +1,153 @@
+;; Test all the f64 operators on major boundary values and all special
```
hjorthjort: File is named `f64-cs.wast`, should be `f64-c.wast`, right?

tests/simple/float/f64_bitwise-cs.wast
```diff
@@ -0,0 +1,147 @@
+;; Test all the f64 bitwise operators on major boundary values and all special
```
hjorthjort: Same here, why is the file named `*-cs.wast`?

tests/simple/float/f64_cmp-cs.wast
```diff
@@ -0,0 +1,76 @@
+;; Test all the f64 comparison operators on major boundary values and all
```
hjorthjort: Same here

wasm.md
```diff
@@ -193,13 +193,14 @@ This allows us to give purely functional semantics to many of the opcodes.
 
 Constants are moved directly to the value stack.
 Function `#unsigned` is called on integers to allow programs to use negative numbers directly.
+**TODO**: Implemented Float in the format of `-nan`, `nan:0x n:hexnum` and `hexfloat`.
```
hjorthjort: Implemented => Implement

wasm.md
```diff
@@ -193,13 +193,14 @@ This allows us to give purely functional semantics to many of the opcodes.
...
 
 ```k
     syntax PlainInstr ::= IValType "." "const" Int
-                        | FValType "." "const" Float
- // ------------------------------------------------
-    rule <k> ITYPE:IValType . const VAL => #chop(< ITYPE > VAL) ... </k>
-    rule <k> FTYPE:FValType . const VAL => < FTYPE > VAL        ... </k>
+                        | FValType "." "const" Number
```
hjorthjort: Is this because a float can given as a number with no `.` in wasm, but it has to have a `.` in K?

wasm.md
```diff
@@ -276,38 +280,44 @@ Comparisons consume two operands and produce a bool, which is an `i32` value.
...
+    rule <k> FTYPE . ROP:FRelOp => FTYPE . ROP C1 C2 ... </k>
+         <valstack> < FTYPE > C2 : < FTYPE > C1 : VALSTACK => VALSTACK </valstack>
+
 ```
 
 ### Conversion Operations
 
 Conversion operators always take a single argument as input and cast it to another type.
-For each element added to `ConvOp`, functions `#convSourceType` and `#convOp` must be defined over it.
+For each element added to `CvtOp`, functions `#convSourceType` and `#CvtOp` must be defined over it.
```
hjorthjort: Any special reason to change the name in this PR? Just wondering, it seems unrelated to the rest of the fixes.

wasm.md
```diff
@@ -276,38 +280,44 @@ Comparisons consume two operands and produce a bool, which is an `i32` value.
...
+         <valstack> < SRCTYPE > C1 : VALSTACK => VALSTACK </valstack>
+      requires #convSourceType(CVTOP) ==K SRCTYPE
+    rule <k> FTYPE:FValType . CVTOP:CvtOp => FTYPE . CVTOP C1  ... </k>
          <valstack> < SRCTYPE > C1 : VALSTACK => VALSTACK </valstack>
-      requires #convSourceType(CONVOP) ==K SRCTYPE
+      requires #convSourceType(CVTOP) ==K SRCTYPE
 
-    syntax IValType ::= #convSourceType ( ConvOp ) [function]
- // ---------------------------------------------------------
+    syntax AValType ::= #convSourceType ( CvtOp ) [function]
```
hjorthjort: Rename `#convSourceType` to `#cvtSourceType`, if you are renaming.

wasm.md
```diff
@@ -276,38 +280,44 @@ Comparisons consume two operands and produce a bool, which is an `i32` value.
...
+                        | FValType "." CvtOp
+ // ----------------------------------------
 
-    syntax Instr ::= IValType "." ConvOp Int
- //                | FValType "." ConvOp Float
+    syntax Instr ::= IValType "." CvtOp Number
+                   | FValType "." CvtOp Number
  // ------------------------------------------
-    rule <k> ITYPE . CONVOP:ConvOp => ITYPE . CONVOP C1  ... </k>
+    rule <k> ITYPE:IValType . CVTOP:CvtOp => ITYPE . CVTOP C1  ... </k>
```
hjorthjort: Why do we need two separate rules here? They look identical to me, except the typing of the `ValType`. But both `IValType` and `FValType` as subsorts of `ValType`, right?

wasm.md
```diff
@@ -328,64 +338,64 @@ Note that we do not need to call `#chop` on the results here.
 ```k
     syntax IBinOp ::= "div_u" | "rem_u"
  // -----------------------------------
-    rule <k> ITYPE . div_u I1 I2 => < ITYPE > I1 /Int I2 ... </k> requires I2 =/=Int 0
-    rule <k> ITYPE . div_u I1 I2 => undefined            ... </k> requires I2  ==Int 0
+    rule <k> ITYPE:IValType . div_u I1 I2 => < ITYPE > I1 /Int I2 ... </k> requires I2 =/=Int 0
```
hjorthjort: I would think the explicit typing is not necessary here, since there is no `div_u` or `div_s` for floats, only `div`, and likewise no `rem_{u|s}`.

wasm.md
```diff
@@ -328,64 +338,64 @@ Note that we do not need to call `#chop` on the results here.
...
 
 `eqz` checks wether its operand is 0.
 
 ```k
-    syntax ITestOp ::= "eqz"
- // ------------------------
-    rule <k> _ . eqz I => < i32 > #bool(I ==Int 0) ... </k>
+    syntax TestOp ::= "eqz"
+ // -----------------------
+    rule <k> _:IValType . eqz I => < i32 > #bool(I ==Int 0) ... </k>
```
hjorthjort: Same here, why do explicit typing if there are no floating point test operators?

wasm.md
```diff
@@ -328,64 +338,64 @@ Note that we do not need to call `#chop` on the results here.
...
-    rule <k> _ . eq I1 I2 => < i32 > #bool(I1 ==Int   I2) ... </k>
-    rule <k> _ . ne I1 I2 => < i32 > #bool(I1 =/=Int  I2) ... </k>
+    rule <k> _:IValType . eq I1 I2 => < i32 > #bool(I1 ==Int  I2) ... </k>
+    rule <k> _:IValType . ne I1 I2 => < i32 > #bool(I1 =/=Int I2) ... </k>
 
     syntax IRelOp ::= "lt_u" | "gt_u" | "lt_s" | "gt_s"
  // ---------------------------------------------------
-    rule <k> _     . lt_u I1 I2 => < i32 > #bool(I1 <Int I2) ... </k>
-    rule <k> _     . gt_u I1 I2 => < i32 > #bool(I1 >Int I2) ... </k>
+    rule <k> _:IValType     . lt_u I1 I2 => < i32 > #bool(I1 <Int I2) ... </k>
```
hjorthjort: Same here, these shouldn't require explicit typing.

wasm.md
```diff
@@ -408,19 +418,19 @@ Careful attention is made for the signed version `shr_s`.
 ```k
     syntax IBinOp ::= "shl" | "shr_u" | "shr_s"
  // -------------------------------------------
-    rule <k> ITYPE . shl   I1 I2 => #chop(< ITYPE > I1 <<Int (I2 %Int #width(ITYPE))) ... </k>
-    rule <k> ITYPE . shr_u I1 I2 =>       < ITYPE > I1 >>Int (I2 %Int #width(ITYPE))  ... </k>
+    rule <k> ITYPE:IValType . shl   I1 I2 => #chop(< ITYPE > I1 <<Int (I2 %Int #width(ITYPE))) ... </k>
```
hjorthjort: And here.

wasm.md
```diff
@@ -408,19 +418,19 @@ Careful attention is made for the signed version `shr_s`.
...
 ```
 
 The rotation operators `rotl` and `rotr` do not have appropriate K builtins, and so are built with a series of shifts.
 
 ```k
     syntax IBinOp ::= "rotl" | "rotr"
  // ---------------------------------
-    rule <k> ITYPE . rotl I1 I2 => #chop(< ITYPE > (I1 <<Int (I2 %Int #width(ITYPE))) +Int (I1 >>Int (#width(ITYPE) -Int (I2 %Int #width(ITYPE))))) ... </k>
-    rule <k> ITYPE . rotr I1 I2 => #chop(< ITYPE > (I1 >>Int (I2 %Int #width(ITYPE))) +Int (I1 <<Int (#width(ITYPE) -Int (I2 %Int #width(ITYPE))))) ... </k>
+    rule <k> ITYPE:IValType . rotl I1 I2 => #chop(< ITYPE > (I1 <<Int (I2 %Int #width(ITYPE))) +Int (I1 >>Int (#width(ITYPE) -Int (I2 %Int #width(ITYPE))))) ... </k>
```
hjorthjort: And here.

wasm.md
```diff
@@ -468,15 +480,128 @@ Wrapping cuts of the 32 most significant bits of an `i64` value.
...
+    syntax CvtOp ::= "extend_i32_u" | "extend_i32_s"
+ // ------------------------------------------------
+    rule <k> i64 . extend_i32_u I:Int => < i64 > I                               ... </k>
+    rule <k> i64 . extend_i32_s I:Int => < i64 > #unsigned(i64, #signed(i32, I)) ... </k>
 
     rule #convSourceType(extend_i32_u) => i32
     rule #convSourceType(extend_i32_s) => i32
 ```
 
+Conversion turns a `int` type value to the corresponding `float` type value.
```
hjorthjort: a => an

And I guess it should be "The closest corresponding `float` type value", right? Many integers don't have an exact float representation.

wasm.md
```diff
@@ -468,15 +480,128 @@ Wrapping cuts of the 32 most significant bits of an `i64` value.
...
+
+```k
+    syntax CvtOp ::= "convert_i32_s" | "convert_i32_u" | "convert_i64_s" | "convert_i64_u"
+ // --------------------------------------------------------------------------------------
+    rule <k> FTYPE:FValType . convert_i32_s I:Int => #round( FTYPE , #signed(i32, I) ) ... </k>
```
hjorthjort: I'm not sure you need the explicit typing here, either, because `#round` only takes `FValType`.

wasm.md
```diff
@@ -468,15 +480,128 @@ Wrapping cuts of the 32 most significant bits of an `i64` value.
...
+    syntax CvtOp ::= "demote_f64" | "promote_f32"
+ // ---------------------------------------------
+    rule <k> f32 . demote_f64  F => #round( f32 , F ) ... </k>
+    rule <k> f64 . promote_f32 F => #round( f64 , F ) ... </k>
+
+    rule #convSourceType(demote_f64)  => f64
+    rule #convSourceType(promote_f32) => f32
+```
+
+Float truncation to int first truncate a float, then convert it to an integer.
```
hjorthjort: You used monospace `float` and `int` above.

wasm.md
```diff
@@ -468,15 +480,128 @@ Wrapping cuts of the 32 most significant bits of an `i64` value.
...
+
+```k
+    syntax Bool ::= #infinityOrNaN ( Float ) [function]
```
hjorthjort: As a predicate, the name `#isInfinityOrNaN` is probably better

wasm.md
```diff
@@ -468,15 +480,128 @@ Wrapping cuts of the 32 most significant bits of an `i64` value.
...
+    rule #convSourceType(trunc_f32_u) => f32
+    rule #convSourceType(trunc_f64_s) => f64
+    rule #convSourceType(trunc_f64_u) => f64
+```
+
+**TODO**: Unimplemented: `inn.reinterpret_fnn`,  `fnn.reinterpret_inn`.
+
+### Floating Point Arithmetic
+
+For the operators that under both sort `IXXOp` and `FXXOp`, we need to give it a `klabel` and define it as a `symbol`.
```
hjorthjort: that under => that is defined under

hjorthjort: sort => sorts

convert.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def main():
+    filename = sys.argv[1]
+    infile = "tests/wasm-tests/test/core/%s" % filename
+    outfile = open("tests/simple/%s-c.%s" % tuple(filename.split(".")), "w")
+    unsupported = ["nan:", "-nan", "reinterpret",
+                   "assert_return_canonical_nan", "assert_return_arithmetic_nan", "assert_invalid", "assert_malformed"]
+    for line in (open(infile).readlines()):
+        if any(x in line for x in unsupported):
```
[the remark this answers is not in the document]
QianyangPeng: it is not safe and could be optimized. But all the conformance test files related to float operators could be handled. They don't have multiline assertions.

tests/simple/float/f64-cs.wast
QianyangPeng: cs means "conformance simplified", as all hexfloats are converted to floats.

tests/simple/conversion.wast
QianyangPeng: in conformance tests float and int conversions are put together, but now float related conformance tests are put separate in a folder called float, so I prefer to keep this for now.

wasm.md
QianyangPeng: Yes. In K if a token is parsed as Int, then it cannot be parsed as Float. So the ambiguity is not allowed and it is pretty annoying.

wasm.md
QianyangPeng: Well, we can have a separate PR for this. The only reason for this renaming is that the Op is named `CvtOp` in the specs

wasm.md
QianyangPeng: fixed

convert.py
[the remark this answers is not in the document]
hjorthjort: Alright, that's good enough. But maybe add a comment clarifying that. This isn't a general solution, just a specific solution for skipping some official conformance tests.

tests/simple/conversion.wast
hjorthjort: Ah, I see.

tests/simple/float/f64-cs.wast
hjorthjort: Aha, so the pre-processor has been run on these files? Maybe put a README in this folder then that explains it.

hjorthjort: And please include in the README how to run the preprocessor, with an example.

QianyangPeng: the preprocessor is just the convert.py file, but yes I'll add document somewhere

QianyangPeng: added some words to convert.py. Prefer not adding a README file here because it is more like a temporary solution before we inplement hexfloat.

hjorthjort: Are we close to having hexfloat working? 

If not, I still prefer some wording somewhere in more prominent documentation. Even if this is meant to be temporary it seems like it could easily become permanent. Having some semi-auto-generated files in the repo that are only distinguished by a naming convention seems to warrant a prominent explanation, especially since the implementer is leaving soon.

PR #147, opened 2019-07-15
Bugfix: `store` wasn't accessing the correct memory

hjorthjort:
Small but devilish bug in the semantics: we didn't look up `CUR` in the configuration. This worked well for all tests of `store` with only one module.

This commit adds a regression test, and fixes the bug.


hjorthjort:
Accidentally pushed another, unrelated commit. Removed now.

PR #149, opened 2019-07-16
Wrap configuration in <wasm> cell

hjorthjort:
This is necessary to make configuration composition work.


ehildenb:
Let's move the `success-haskell.out` file.

PR #157, opened 2019-07-17
Update makefile with parameters (for inclusion in other repo)

hjorthjort:
Ewasm imports KWasm, and calls `make` recursively. These parameters means we get more granular control over where definitions end up.

Not sure I understood how to do this correctly, I don't see an improvement in build times. How can I make sure I'm not rebuilding the definitions for Wasm, for example?


hjorthjort:
See the Ewasm Makefile here: https://github.com/kframework/ewasm-semantics/blob/initial/Makefile

hjorthjort:
Note to self: seems the builds still don't get put into the directories named after `kompiled_dir_name`, probably needs a flag or something to `kompile`, the rule changes don't really redirect the build files.

PR #168, opened 2019-07-26
the framework of binary module implementation

QianyangPeng:
This PR mainly does the following:
1. renamed wasm.md to core.md, and renamed module WASM to WASM-CORE
2. add a file binary.md and a module WASM-BINARY
3. add a file wasm.md and a module WASM that just imports WASM-CORE and WASM-BINARY
4. implemented the rough strcture of loading a binary module. Currently only custom sections are implemented, whose functionality is nothing but just skip the whole section. (according to the specs)

The current implementation is not a very efficient way to do the work, but I do not have a better solution for now. If it looks good enough we can continue implementing the rest of sections.


QianyangPeng:
it is closed because a more functional approach is preferred. But in order to do that, we have the following difficult points:
1. We prefer to manipulate the raw string, but in the raw string every byte is of a variable length. Some of them has length 3 ( e.g. “\97” ) and some of them has length 1 ( e.g. “a” ). So its hard to manage it without doing any encoding/decoding first.
2. The length of section is encoded into a LEB128^{28} length. We need to decode it first.

It may be just a too tricky task and doesn't really worth the effort.

PR #178, opened 2019-08-11
Text format

hjorthjort:
Introduce a separate module for the text format.

Instead of having all these rules in the core syntax, we can keep them in a separate module. This both makes the core semantics smaller, more like the official spec, and a smaller compile target for the binary format.

This is not a full separation of text format and core semantics. The following remains:
1a. Getting rid of any mentions identifiers in the definitions in the core semantics, ideally by making all the definitions more abstract in the core semantics, and only giving their text format, with identifiers and all, in the text semantics. This can be achieved since this PR splits the definitions of `Index`, and all lookups pass through the `#ContextLookup` function.
1b. Getting rid of identifiers on branches and labels in the core semantics (requires some way to look up the correct branching index).
2. Only using integers in the core semantics and getting rid of the `#ContextLookup`. 
  2a. Move all definitions (functions, memories, etc.) with identifiers to the text format.
  2b. Get rid of `#ContextLookup` altogether: remove it from types lookups, etc.
3. Getting rid of all the `<*Ids>` cells in the core configuration and use configuration composition to add identifiers.

I'd say 1a. is really nice, and I have actually implemented it, and have an upcoming PR for it. 1b could be achieved by looking through the `<k>` cell, counting labels until you get to the label you are looking for. Not super elegant, but gets the job done. 2. is also a nice improvement, and because `#ContextLookup` is an inelegant dependency on the text format in the core semantics. 3. is perhaps a bit overkill.

### DIFF
--- a/wasm-text.md
+++ b/wasm-text.md
@@ -17,7 +17,7 @@ module WASM-TEXT
 Folded Instructions
 -------------------
 
-Folded Instructions (`FoldedInstr`): Folded Instructions describes the rules of desugaring plain instructions and block instructions.
+Folded Instructions describes the rules of desugaring plain instructions and block instructions.
 
 ```k
     syntax Instr ::= FoldedInstr
--- a/wasm-text.md
+++ b/wasm-text.md
@@ -42,6 +42,77 @@ Folded Instructions describes the rules of desugaring plain instructions and blo
     rule <k> ( loop OID:OptionalId TDECLS:TypeDecls IS ) => loop OID TDECLS IS end ... </k>
 ```
 
+Block Instructions and Control
+------------------
+
+In the abstract Wasm syntax, indices are always integers.
+In the text format, we extend indices to incorporate identifiers.
+
+```k
+    syntax Index ::= Identifier
+ // ---------------------------
+```
+
+First, we enable context lookups with identifiers.
+
+```k
+    rule #ContextLookup(IDS:Map, ID:Identifier) => {IDS [ ID ]}:>Int
+      requires ID in_keys(IDS)
+```
+
+Then, we define how to branch to identifiers.
+
+```k
+    rule <k> br ID:Identifier ~> label ID  [ TYPES ] { IS } VALSTACK' => IS ... </k>
+         <valstack> VALSTACK => #take(TYPES, VALSTACK) ++ VALSTACK' </valstack>
+    rule <k> br ID:Identifier ~> label ID' [ TYPES ] { IS } VALSTACK' => br ID ... </k>
+      requires ID =/=K ID'
+```
+
+Structured control instructions are used to control the program flow.
+
+```k
+    syntax Instr ::= BlockInstr
+ // ---------------------------
+
+    syntax BlockInstr ::= "block" OptionalId TypeDecls Instrs "end" OptionalId
+ // --------------------------------------------------------------------------
+    rule <k> block OID:OptionalId TDECLS IS end OID':OptionalId => IS ~> label OID gatherTypes(result, TDECLS) { .Instrs } VALSTACK ... </k>
+         <valstack> VALSTACK => .ValStack </valstack>
+      requires OID ==K OID'
+        orBool notBool isIdentifier(OID')
+```
+
+Finally, we have the conditional and loop instructions.
+
+```k
+    syntax BlockInstr ::= "if" OptionalId TypeDecls Instrs "else" OptionalId Instrs "end" OptionalId
+                        | "if" OptionalId TypeDecls Instrs                          "end" OptionalId
+ // ------------------------------------------------------------------------------------------------
+    rule <k> if OID:OptionalId TDECLS:TypeDecls IS                         end OID'':OptionalId => if OID TDECLS IS else OID .Instrs end OID ... </k>
+      requires OID ==K OID''
+        orBool notBool isIdentifier(OID'')
+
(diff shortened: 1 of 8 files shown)


wasm-text.md
```diff
@@ -0,0 +1,278 @@
...
+In the text format, we extend indices to incorporate identifiers.
+
+```k
+    syntax Index ::= Identifier
+ // ---------------------------
+```
+
+We enable context lookups with identifiers.
+
+```k
```
ehildenb: Let's smash these codeblocks together and put the comment above them.

wasm-text.md
```diff
@@ -0,0 +1,278 @@
...
+
+Block Instructions
+------------------
+
+In the text format, block instructions can have identifiers attached to them, and branch instructions can refer to these identifiers.
+First, we allow identifiers on labels.
+
+```k
+    syntax Label ::= "label" Identifier VecType "{" Instrs "}" ValStack
+ // --------------------------------------------------------------------
```
ehildenb: extra `-`

wasm-text.md
```diff
@@ -0,0 +1,278 @@
...
+Then, we define how to branch to identifiers, and how the branches using integers still match on named labels.
+
+```k
+    rule <k> br 0     ~> ( label _:Identifier [ TYPES ] { IS } VALSTACK' => label [ TYPES ] { IS } VALSTACK' ) ... </k>
+
+    rule <k> br ID:Identifier ~> label ID  [ TYPES ] { IS } VALSTACK' => IS ... </k>
+         <valstack> VALSTACK => #take(TYPES, VALSTACK) ++ VALSTACK' </valstack>
+    rule <k> br ID:Identifier ~> label [ TYPES ] { IS } VALSTACK' => br ID ... </k>
+    rule <k> br ID:Identifier ~> label ID' [ TYPES ] { IS } VALSTACK' => br ID ... </k>
+      requires ID =/=K ID'
```
ehildenb: formatting here is a little weird

ehildenb: Also, `br` is semantics, it feels weird to duplicate these here. Are we sure there isn't a way to desugar this to use the integer index? Otherwise, let's move back over to the semantics for now.

wasm-text.md
```diff
@@ -0,0 +1,278 @@
...
+      requires ID ==K OID'
+        orBool notBool isIdentifier(OID')
+
+    syntax BlockInstr ::= "loop" Identifier TypeDecls Instrs "end" OptionalId
+ // -------------------------------------------------------------------------
+    rule <k> loop ID:Identifier TDECLS:TypeDecls IS end OID':OptionalId => IS ~> label ID gatherTypes(result, TDECLS) { loop ID TDECLS IS end } VALSTACK ... </k>
+         <valstack> VALSTACK => .ValStack </valstack>
+      requires ID ==K OID'
+        orBool notBool isIdentifier(OID')
+```
```
ehildenb: Is desugaring `block` and `loop` to `labels` semantics or text format?

wasm-text.md
```diff
@@ -0,0 +1,278 @@
...
+         <valstack> < i32 > VAL : VALSTACK => VALSTACK </valstack>
+      requires VAL =/=Int 0
+       andBool ( ID ==K OID'  orBool notBool isIdentifier(OID')  )
+       andBool ( ID ==K OID'' orBool notBool isIdentifier(OID'') )
+
+    rule <k> if ID:Identifier TDECLS:TypeDecls IS else OID':OptionalId IS' end OID'':OptionalId => IS' ~> label ID gatherTypes(result, TDECLS) { .Instrs } VALSTACK ... </k>
+         <valstack> < i32 > VAL : VALSTACK => VALSTACK </valstack>
+      requires VAL ==Int 0
+       andBool ( ID ==K OID'  orBool notBool isIdentifier(OID')  )
+       andBool ( ID ==K OID'' orBool notBool isIdentifier(OID'') )
```
ehildenb: These also feel like semantics to me.

wasm-text.md
hjorthjort: Done

wasm-text.md
hjorthjort: Done

wasm-text.md
hjorthjort: Move backed to core semantics.

wasm-text.md
hjorthjort: Semantics, moved back.

wasm-text.md
hjorthjort: Moved back.

wasm-text.md
```diff
@@ -0,0 +1,187 @@
...
+
+```k
+    syntax MemorySpec ::= "(" "data" DataString ")"
+ // -----------------------------------------------
+    rule <k> ( memory ( data DS ) ) => ( memory #freshId(NEXTID) (data DS) ) ... </k>
+         <nextFreshId> NEXTID => NEXTID +Int 1 </nextFreshId>
+
+    rule <k> ( memory ID:Identifier ( data DS ) )
+          =>  memory { ID #lengthDataPages(DS) #lengthDataPages(DS) }
+          ~> ( data ID (i32.const 0) DS ) ... </k>
```
ehildenb: formatting `...` and `</k>` on separate lines.

hjorthjort: Fixed

PR #183, opened 2019-08-14
Text format definitons

hjorthjort:
This is part 2 of the migration described in #178

### DIFF
--- a/wasm-text.md
+++ b/wasm-text.md
@@ -59,6 +59,57 @@ Another type of folded instruction is control flow blocks wrapped in parentheses
     rule <k> ( loop ID:Identifier TDECLS:TypeDecls IS ) => loop ID TDECLS IS end ... </k>
 ```
 
+Block Instructions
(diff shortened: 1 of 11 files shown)


wasm-text.md
```diff
@@ -133,8 +213,145 @@ In the text format, it is also allowed to have a conditional without the `else`
        andBool ( ID ==K OID'' orBool notBool isIdentifier(OID'') )
 ```
 
-Memory and Tables
------------------
+Definitions
+-----------
+
+TODO: It's pretty ugly to access `NEXTIDX` the way we are doing here, ideally it should be more functional.
+
```
hjorthjort: Probably nicer to either put `#saveIdentifer` *before* the defintion (and avoid the `-Int 1` part), or just get rid of this and inline the identifier updates.

ehildenb:
Closing for now, though it would be nice to have done at some point.

PR #193, opened 2019-08-29
Official tests

hjorthjort:
# Run conformance tests
Grabs all the official tests (as symlinks), split them into two groups:
- Supported: those that terminate with an empty K cell
- Unsupported: those that don't

Puts the `spec` repo under `deps`. Don't know if that makes sense or not, I think it does.

Also adds the `make` target `test-official` which tests those that are supported. Good for regression testing.

# Bug fixes and regression tests

- Adds the `assert_exhausted` assertion, which we were missing.
- Fixes two branching issues
  1. When calling into a function, a block is added right inside the frame, so it's always possible to break out to the end of the function body
  2. It's possible for us to encounter an `if` block that has been unfolded and turned into an instruction, when the branch occurs in the condition (as in `(if (br 0) then nop else nop)`. We can deal with that by having `br` consume all kinds of `Instr`, not just `PlainInstr`, which I don't know why we were restricting ourselves to.


hjorthjort:
Looks like it's the upstream Rust build error with `MaybeUninit` that's causing this to fail. Tests are passing on my end.

Should we, in general, re-run the `install-rust` script regularly?

.gitmodules
```diff
@@ -3,7 +3,7 @@
 	url = https://github.com/kframework/k
 	ignore = untracked
 [submodule "tests/wasm-tests"]
-	path = tests/wasm-tests
+	path = deps/spec
```
ehildenb: Either (1) don't do this, or (2) also change the submodule name (line above this).

I would lean towards (1), we're only using the spec repo as tests, not as dependencies.

Makefile
```diff
@@ -198,6 +205,11 @@ test-prove: $(proof_tests:=.prove)
 
 test-klab-prove: $(quick_proof_tests:=.klab-prove)
 
+### Official test suite
+
+official_tests:=$(wildcard tests/official/supported/*.wast)
+test-official: $(official_tests:=.run-term)
```
ehildenb: This is conceptually duplicate of the `Conformance` block above.

Instead, rename `parse-conformance` to `test-parse-conformance`, and then add targets `test-conformance-supported` and `test-conformance-unsupported` (so it's easy to run either). Also add `test-conformance` which runs both `test-conformance-supported` and `test-conformance-unsupported`.

tests/official/look_for_supported.sh
```diff
@@ -0,0 +1,19 @@
...
+    if [ $? -eq 0 ]
+    then
+        file_supp=$(echo $file | sed -s "s/unsupported/supported/")
+        git mv $file $file_supp
+        git commit -m "TEST: $shortname"
+        notify-send "Success: $shortname"
+    else
+        notify-send "Failed: $shortname"
+    fi
+done
```
ehildenb: This can go away if you have `test-conformance-unsupported` in the Makefile.

ehildenb:
@hjorthjort the upstream issue is fixed in this PR https://github.com/kframework/wasm-semantics/pull/194#pullrequestreview-281749373

Once that is merged this should be good, but I still would prefer we migrate to a files-based exclusion method for the unsupported tests, and then not bother with the symlinks, just run the tests directly (as `parse-conformance` does).

So, for example, we have a list of failing tests for the LLVM backend: https://github.com/kframework/evm-semantics/blob/master/tests/failing.llvm (edited) 
Then we filter those out of the official test suite: https://github.com/kframework/evm-semantics/blob/d9629df967890587d3a756b3d5786f1ccc1f1283/Makefile#L480

Makefile
hjorthjort: Yes, but it's slow. I thought we were aiming to have a list of unsupported files, so better to do like in KEVM and maintain those files. Then the script makes sense since it updates those files. It could be made a make target, but I'd prefer if the `make` target doesn't modify committed files.

.gitmodules
```diff
@@ -2,7 +2,7 @@
 	path = deps/k
 	url = https://github.com/kframework/k
 	ignore = untracked
-[submodule "tests/wasm-tests"]
+[submodule "test/wasm-tests"]
```
ehildenb: test => tests

Makefile
```diff
@@ -20,6 +20,7 @@ export LUA_PATH
         defn defn-ocaml defn-java defn-haskell defn-llvm \
         build build-ocaml defn-haskell build-haskell build-llvm \
         test test-execution test-simple test-prove test-klab-prove \
+				test-conformance test-conformance-parse test-conformance-supported \
```
ehildenb: let's use the same spacing as the line before

Makefile
```diff
@@ -183,8 +190,15 @@ test-simple-float: $(simple_tests-float:=.frun)
 ### Conformance Tests
 
 conformance_tests:=$(wildcard tests/wasm-tests/test/core/*.wast)
+unsupported_conformance_tests:=$(patsubst %, tests/wasm-tests/test/core/%, $(shell cat tests/official/unsupported.txt))
+unparseable_conformance_tests:=$(patsubst %, tests/wasm-tests/test/core/%, $(shell cat tests/official/unparseable.txt))
+parseable_conformance_tests:=$(filter-out $(unparseable_conformance_tests), $(conformance_tests))
+supported_conformance_tests:=$(filter-out $(unsupported_conformance_tests), $(parseable_conformance_tests))
+
+test-conformance-parse: $(parseable_conformance_tests:=.parse)
+test-conformance-supported: $(supported_conformance_tests:=.run-term)
```
ehildenb: I would also add `test-conformance-unsupported` (but don't add as subtarget to `test-conformance` target), for ease of saying "what should I work on next?" Up to you.

hjorthjort: Sure, can just be a `cat` of the files keeping track.

Makefile
hjorthjort: whoops, looks like tabs vs. spaces. fixed

Makefile
hjorthjort: Won't fix for now.

PR #198, opened 2019-08-31
Build updates for quieter more reliable build

ehildenb:
-   Better dependencies on generated directories.
-   Quieter tangling and kompiling.
-   Switch to LLVM backend default in runner script/Makefile.
-   Allow passing options to `kompile` when using Makefile.
-   Guard against performance regressions in the tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,4 @@
 .krun*
 .kompile*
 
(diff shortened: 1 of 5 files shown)


ehildenb:
lol, thanks auto-merger

PR #216, opened 2019-09-17
Improved lemmas and memory-spec

hjorthjort:
For some reason the prover can't infer that when you have the stack
```
#chop( <i32> EA) : #chop (<i32> EA) : STACK
```
and you pass these as the IDX variable through the rewrites
```
    rule <k> ITYPE . LOP:LoadOp               => ITYPE . LOP  IDX                          ... </k>
         <valstack> < i32 > IDX : VALSTACK    => VALSTACK </valstack>
```
and
```
    rule <k> ITYPE . SOP:StoreOp               => ITYPE . SOP  IDX                          VAL ... </k>
         <valstack> < ITYPE > VAL : < i32 > IDX : VALSTACK => VALSTACK </valstack>
```

that the resulting IDXs should have the same value. I.e., you get a `load` and a `store` that consume the addresses on the stack, but can't see that they would be pointing to the same location.

I tried adding `smtlib(chop)` to the `#chop` function, to no avail (likely because `IVal` is still not a function, so the prover can't reason about the equality of `<i32>` EA to itself when passed through a function).

One thing that did work was removing the `[concrete]` annotation to chop and letting it pass down to the underlying `modInt`. Then we can use some straightforward modular arithmetic lemmas to reason about `#chop` instead of the dedicated rules.

This was mostly an exercise in trying to get rid of the `#inUnsignedRange` side condition on `memory-spec.k`. I think it's nice if we can do without it, but I'd like to get a second opinion.


hjorthjort:
Basically, it comes down to whether we want to include the `#in(Un)signedRange` for all our proofs and just let the `#chop` get stripped off, or if we want to allow integers that are bigger/smaller than the range and make sure we can reason about it correctly.

hjorthjort:
Found something interesting in the spec. It says the text format _and_ the binary format disallow integers outside the signed and unsigned range as `i32.const`. 

But integers in the _signed_ range are allowed. When I try to replace `#inUnsignedRange` in the memory proof with `#inUnsginedRange(...) orBool #inSignedRange(...)` it doesn't go through.

The issue is that the prover doesn't split on `orBool` side conditions. If it did, it should be able to work this out fine. Because if I run the proofs with either `#inUnsignedRange` or `#inSignedRange` as a side condition, the proof goes through.

I think that's an indication that we should go with the approach above, or at least something similar. That will let Z3 reason about the simple arithmetic of equalities in `#chop`.

kwasm-lemmas.md
```diff
@@ -13,24 +13,28 @@ Basic arithmetic
...
 
     rule (X +Int (Y *Int N)) modInt N => X modInt N
       requires N >=Int 1
+      [simplification]
     rule (X +Int (Y *Int N)) /Int N   => (X /Int N) +Int Y
       requires N =/=Int 0
+      [simplification]
+
+    rule X modInt N => X
+      requires X <=Int 0
```
ehildenb: Should this be `X >=Int 0`

kwasm-lemmas.md
```diff
@@ -13,24 +13,28 @@ Basic arithmetic
 ----------------
 
 ```k
-    rule (0 +Int X) => X
-    rule (X +Int 0) => X
+    rule (0 +Int X) => X [simplification]
+    rule (X +Int 0) => X [simplification]
 
     rule (X +Int (Y *Int N)) modInt N => X modInt N
       requires N >=Int 1
```
ehildenb: Can this be generalized to `N =/=Int 0` instead? It then looks more similar to the rule for `/Int` below.

Though, not sure what `modInt` with a negative number means...

hjorthjort: Yeah, usually in modular arithmetic the modulus is assumed to be a positive integer, and different languages implement the negative case differently. In Python `12 % -5 == -3`, in Java `12 % -5 == 2`, and in Haskell
```
12 `mod` (-5) == -3
12 `rem (-5) == 2
```
Either way, this should hold, but I don't want to get too general due to the weirdness of mod/rem.

hjorthjort: Finally, in z3 the result of mod is always non-negative, it seems. The following is unsat:

```z3
(declare-const x   Int)
(declare-const y   Int)
(declare-const res Int)

(assert (< y 0))
(assert (< res 0 ))
(assert (= res (mod x y)))
```

hjorthjort: We could of course make the condition `N =/=Int 0 andBool notBool (N <Int 0)` or something, if that will make resoning easier for the prover.

hjorthjort: And for the sake of establishing trust in the trusted codebase, this verifies the claim in z3, starting from the assumption that `mod` distributes.

```
(declare-const x   Int)
(declare-const y   Int)
(declare-const n   Int)

(assert (forall ((i Int) (j Int) (k Int)) (= (mod (+ i j) k) (+ (mod i k) (mod j k))))) ; mod distribtues over addition.

(assert (< 0 n))
(assert (not (= (mod (+ x (* y n)) n) (mod x n))))

(check-sat)
```
Gives `unsat`.

kwasm-lemmas.md
hjorthjort: Good catch. I'll add a regression tests to `simple-arithmetic-spec`.

hjorthjort:
Added regression testing capabilities (ba8356a) for the trusted computing base. What do you think? Not sure if it's elegant/overkill, but it's nice if we can make sure the prover can't do certain (bad) things, it's good for sanity checking, too

Makefile
```diff
@@ -176,6 +177,13 @@ tests/%.parse: tests/%
 tests/%.prove: tests/%
 	$(TEST) prove --backend $(TEST_SYMBOLIC_BACKEND) $< --format-failures --def-module $(KPROVE_MODULE)
 
+tests/%.cannot-prove: tests/%
+	@echo "$<"
+	@failed=0 \
+	  ; $(TEST) prove --backend $(TEST_SYMBOLIC_BACKEND) $< --format-failures --def-module $(KPROVE_MODULE) --boundary-cells k 2> /dev/null \
+	  || failed=1 \
+	  ; test $$failed = 1
```
ehildenb: Let's instead test failed tests via an expected output file, with the failed proof state.

EG:

```
$(TEST) prove --backend $(TEST_SYMBOLIC_BACKEND) $< --format-failures --def-module $(KPROVE_MODULE) --boundary-cells k > $<.out \
    || $(CHECK) $<.expected $<.out
```

Then add expected output files (by running the test then copying the expected output file into place).

ehildenb: If `$(CHECK)` isn't defined yet (not sure), it should be `CHECK := git --no-pager diff --no-index`

tests/bad-proofs/mod-on-negative-spec.k
```diff
@@ -0,0 +1,10 @@
+requires "kwasm-lemmas.k"
+
+module MOD-ON-NEGATIVE-SPEC
+    imports KWASM-LEMMAS
+
+    rule <k> X modInt N => X ... </k>
+      requires X <=Int 0
+       andBool 0  <Int N
+
+endmodule
```
ehildenb: Would recommend taking at least one semantic step in all proofs to avoid weird corner cases in the java backend.

Something as simple as:

```
syntax Stmt ::= doStep ( Int )
rule <k> doStep ( I ) => I ... </k>
```

Makefile
hjorthjort: Done

tests/bad-proofs/mod-on-negative-spec.k
hjorthjort: Trying to get this to work ...

Added  to test.md:
```
    syntax KItem ::= "doStep" "(" Int ")" | "doneStep" "(" Int ")"
 // --------------------------------------------------------------
    rule <k> doStep ( I ) => doneStep ( I ) ... </k>
```

and the proof spec is now:

```
    rule <k> doStep ( X modInt N ) => doneStep ( X ) ... </k>
      requires X <=Int 0
       andBool 0  <Int N
```

The prover fails, seemingly because it fails to match on the initial configuration, ` Subject: doStep(_modInt_(X_918:Int,, N_917:Int))`

PR #224, opened 2019-09-27
More core test support

hjorthjort:


hjorthjort:
This ad-hoc attempt isn't a great solution. We should instead make sure that when we import from one of the special modules, we introduce a tainted value (a memory, a global, a function) so that when it gets touched, the special trap shows up.

PR #225, opened 2019-09-27
Parse more tests

hjorthjort:


wasm.md
```diff
@@ -1173,9 +1173,9 @@ This is not optional.
 The offset can either be specified explicitly with the `offset` key word, or be a single instruction.
 
 ```k
-    syntax Offset ::= "(" "offset" Instr ")"
```
ehildenb: Should the comment above be updated?

ehildenb: (Instructions instead of "a single instruction")

ehildenb: Also, why are we allowing multiple instructions, does the spec allow it?

hjorthjort: The answer is: the grammar allows it, but it makes the module invalid. So we need this to parse the core tests, but since we just ignore the `assert_invalid` modules, we may assume in the rules that it's always a single instruction.

PR #236, opened 2019-10-23
prove memory spec using haskell backend

yzhang90:


hjorthjort:
Nice! Add a line to the Jenkins file also to make sure this is tested in CI.
https://github.com/kframework/wasm-semantics/blob/master/Jenkinsfile#L117

yzhang90:
The haskell backend wants to include the memory proof in the integration test of their PR. So they want to merge the change to the semantics first --> they will merge the corresponding PR --> we will update the haskellbackend in KWASM and add it to integration test of KWASM.

@ttuegel please confirm the above workflow.

PR #243, opened 2019-10-31
Polkadot updates - decrease state splitting

ehildenb:


wasm.md
```diff
@@ -1452,42 +1465,36 @@ Imports must appear before any allocations in a module, due to validation.
...
- // -------------------------------------------------------
-    rule #initialMap ()
-      => "typeDecls" |-> .Defns
-         "imports"   |-> .Defns
-         "func/glob" |-> .Defns
-         "allocs"    |-> .Defns
-         "exports"   |-> .Defns
-         "inits"     |-> .Defns
-         "start"     |-> .Defns
+    syntax ModuleDecl ::= sortedModule ( id: OptionalId, types: Defns, importDefns: Defns, funcsGlobals: Defns, allocs: Defns, exports: Defns, inits: Defns, start: Defns )
```
hjorthjort: Would it help to refine the types of `Defns` further, into FuncDefns, etc? If so I guess we should make sure  any `Defns` list is read-only to get proper subtyping.

PR #251, opened 2019-11-06
Dockerfile: Install z3-4.8.6

ttuegel:


hjorthjort:
Do the newer versions not support the `timeout` to `using-params`? Should we switch to using the command line flags instead?

From `z3 -h` on version 4.8.5:

```
Resources:
  -T:timeout  set the timeout (in seconds).
  -t:timeout  set the soft timeout (in milli seconds). It only kills the current query.
```

hjorthjort:
@ttuegel Is this PR still of interest? Any idea why the checks are failing?

ttuegel:
No, it's fine to close this, but we should keep in mind that the frontend and backend will both be slower with the older version.

hjorthjort:
@ttuegel Any estimate how much slower?

PR #257, opened 2019-11-15
Reverse bytes spec

hjorthjort:


kwasm-lemmas.md
```diff
@@ -82,6 +202,9 @@ Conversely, setting an index in a map to a value `VAL` and then retrieving the v
 
 ```k
     rule #set(BMAP, IDX, #get(BMAP, IDX)) => BMAP [smt-lemma]
+
+    rule #get(#set(BMAP, IDX,  VAL), IDX) => VAL  [smt-lemma]
+    rule IDX ==Int IDX' impliesBool #get(#set(BMAP, IDX', VAL), IDX) ==Int #get(BMAP, IDX) => true [smt-lemma]
```
hjorthjort: This is a mistake. It gets the final implication to pass, but I really don't understand why.

kwasm-lemmas.md
```diff
@@ -20,13 +20,157 @@ Basic logic
...
+Assume m = n * k for some k > 0.
+x = m * q + r, for a unique q and r s.t. 0 <= r < m
+(x mod m) mod n
+ = r mod n
+ = n * ( k * q) + r mod n
+ = m * q + r mod n
+ = x mod n
+```
+
+#### Modulus Over Addition and Division
```
ehildenb: There isn't any division in this section.

kwasm-lemmas.md
```diff
@@ -20,13 +20,157 @@ Basic logic
...
+  = r + y mod n
+  = (x mod m + y) mod n
+```
+
+#### Bit Shifting
+
+We want Z3 to understand what a bit-shift is.
+
+```k
+    rule ( X >>Int N)          => 0 requires X <Int 2 ^Int N [simplification]
```
ehildenb:
```suggestion
    rule (X >>Int N)          => 0 requires X <Int 2 ^Int N [simplification]
```

kwasm-lemmas.md
```diff
@@ -20,13 +20,157 @@ Basic logic
...
+    rule ( X <<Int N) modInt M => 0 requires M <Int 2 ^Int N [simplification]
```
ehildenb:
```suggestion
    rule (X <<Int N) modInt M => 0 requires M <Int 2 ^Int N [simplification]
```

kwasm-lemmas.md
```diff
@@ -20,13 +20,157 @@ Basic logic
...
 ```
 
+`modInt` selects the least non-negative representative of a congruence class modulo `N`.
+
+```k
+    rule (X modInt M) modInt N => X modInt M
+      requires M >Int 0
+       andBool M <=Int N
+      [simplification]
+```
```
hjorthjort: Specific cae of the above. Add smt-lemma?
```
rule A ==Int X modInt M impliesBool 0 <=Int A andBool A <Int M => true [smt-lemma]
```

hjorthjort: Try adding smt-lemma and removing above lemma, still pass?
Investigate why z3 doesn't get this

hjorthjort: For thesis: document workflow

hjorthjort: Answer question: Why can't we just always posit all lemmas we know? Example: Associativity.
Reason: rewrite-based, infitine looping rewrites. Aiming for a normal form.

hjorthjort: (X + Y) mod N = (X mod N + Y mod N) mod N would solve a lot.

kwasm-lemmas.md
```diff
@@ -20,13 +20,157 @@ Basic logic
...
+
+Proof:
+
+
```
hjorthjort:
```suggestion
```

kwasm-lemmas.md
hjorthjort: Double check: all side conditions guarantee M and N are not 0

kwasm-lemmas.md
```diff
@@ -20,13 +20,156 @@ Basic logic
...
+    rule (X >>Int N) >>Int M => X >>Int (N +Int M) [simplification]
+    rule (X <<Int N) <<Int M => X <<Int (N +Int M) [simplification]
+
+    // The Haskell and Java backend accept negative shifts (the LLVM and OCaml backends do not).
+    // So removing the side conditions and keeping one of each rule here could give faster symbolic execution.
+    rule (X <<Int N) >>Int M => X <<Int (N -Int M)   requires          N >=Int M  [simplification]
+    rule (X <<Int N) >>Int M => X >>Int (M -Int N)   requires notBool (N >=Int M) [simplification]
+    rule (X >>Int N) <<Int M => X >>Int (N -Int M)   requires          N >=Int M  [simplification]
+    rule (X >>Int N) <<Int M => X <<Int (M -Int N)   requires notBool (N >=Int M) [simplification]
```
hjorthjort: Last 2 are unsound

hjorthjort: Can be expressed as a "blanking operation"
Shifting 8 bites right and then 8 bits left is blanking the leat 8 bits.

kwasm-lemmas.md
```diff
@@ -71,6 +221,13 @@ We want to make the variant explicit, so we introduce the following helper, whic
       requires notBool isInt(I)
 ```
 
+We can now state the intended semantics of `#get`.
+
+```k
+    rule 0              <=Int #get(BM, ADDR) => true requires #isByteMap(BM) [simplification]
+    rule #get(BM, ADDR)  <Int 256            => true requires #isByteMap(BM) [simplification]
```
hjorthjort: Do these help? Try removing

hjorthjort: Can we get a log to find out what applies?
Can I automatically detect rule applications? Good for report/general grokking.

kwasm-lemmas.md
```diff
@@ -20,13 +20,154 @@ Basic logic
...
+    // The Haskell and Java backend accept negative shifts (the LLVM and OCaml backends do not).
+    // So removing the side conditions and keeping one of each rule here could give faster symbolic execution.
+    rule (X <<Int N) >>Int M => X <<Int (N -Int M)   requires          N >=Int M  [simplification]
+    rule (X <<Int N) >>Int M => X >>Int (M -Int N)   requires notBool (N >=Int M) [simplification]
+```
+
+The following rules decrease the modulus by rearranging it around a shift.
+
+```k
+    rule (X          modInt POW) >>Int N => (X >>Int N) modInt (POW /Int (2 ^Int N))            requires POW modInt (2 ^Int N) ==Int 0 [simplification]
```
yzhang90: Can I change the formula to 
```
   rule ( X  modInt POW) >>Int N => ( X >>Int N) modInt  (POW >>Int N) requires ...
```

kwasm-lemmas.md
```diff
@@ -20,13 +20,154 @@ Basic logic
...
+    rule (X +Int (Y <<Int N)) modInt POW => X modInt POW   requires (2 ^Int N) modInt POW ==Int 0 [simplification]
+    rule ((Y <<Int N) +Int X) modInt POW => X modInt POW   requires (2 ^Int N) modInt POW ==Int 0 [simplification]
+```
+
+Proof: These follow from the fact that shifting left by `n` bits is simply multiplying by `2^n`, and from previously proven rules of modular arithmetic.
+
+The next rules consider (some) cases where integers with disjoint 1-bits are added together.
+
+```k
+    rule (X +Int Y) >>Int N => (Y >>Int N)
```
yzhang90: is this correct? if X = 3, Y = 7, N = 2,  (X + Y) >> N = 10 / 4 = 2, Y >> N = 7 / 4 = 1.  
In fact, if the side condition is X <Int 2 ^Int (N -Int 1), you can prove the addition causes no carriers.

kwasm-lemmas.md
hjorthjort: That would be nicer, but Z3 doesn't seem to know about in K, so this won't work.
Shift only works on bitvectors.

Z3 does support reasoning about bitvectors <https://rise4fun.com/z3/tutorial/guide> (Section 6). But AFAIKT, there is no support for this in K.

kwasm-lemmas.md
hjorthjort: Good catch, you're 100% right.
Turns out this lemma isn't needed for the proof to go through.

hjorthjort:
@yzhang90 Cleaning this up now. One question:
For upstreaming, do we prefer:
```
X modInt N => <something>
  requires N =/=Int 0
```
or
```
X modInt N => <something>
  requires N >Int 0
```
?

In K, `X modInt N ==Int X modInt (-N)`, so both should hold up. But the first case introduces some cognitive complexity, since modulus is usually defined (and most theorems are written for) the non-negative case.

I tend to prefer `N >Int 0` for this reason (easier to trust), but is there a good reason to be more general? Wasm should never do modulus of a negative number, and I don't really see the use case for that.

yzhang90:
> @yzhang90 Cleaning this up now. One question:
> For upstreaming, do we prefer:
> 
> ```
> X modInt N => <something>
>   requires N =/=Int 0
> ```
> 
> or
> 
> ```
> X modInt N => <something>
>   requires N >Int 0
> ```
> 
> ?
> 
> In K, `X modInt N ==Int X modInt (-N)`, so both should hold up. But the first case introduces some cognitive complexity, since modulus is usually defined (and most theorems are written for) the non-negative case.
> 
> I tend to prefer `N >Int 0` for this reason (easier to trust), but is there a good reason to be more general? Wasm should never do modulus of a negative number, and I don't really see the use case for that.

I also prefer `N >Int 0`. Currently there is no semantics that uses modulo negative numbers. We can always update the lemma when we meet such cases.

kwasm-lemmas.md
```diff
@@ -352,3 +634,92 @@ A module of shorthand commands for the WRC20 module.
 ```k
 endmodule
 ```
+
+TODO: Upstream these, but don't commit them with the WRC20 proof (they aren't needed).
+
+```k
+module GENERAL-ARITHMETIC-LEMMAS
+    imports WASM-TEXT
```
yzhang90: is `imports WASM-TEXT` needed?

kwasm-lemmas.md
```diff
@@ -20,13 +20,198 @@ Basic logic
...
+`X modInt N` is undefined for `N ==Int 0`, so we must take care to check that `N =/=Int 0`.
+At the same time, we don't want to impose unnecessary side-conditions, so the non-zero condition can be implied by the other conditions.
+For example, `X modInt N ==Int Y` implies `N =/=Int 0`, because only non-zero `N` could have the right-hand side resolve to an `Int`.
+For simplicity, we impose that `N >Int 0`.
+Not however that K defines `X modInt N ==Int X modInt (-N)`.
+
+#### Rules for Expressions With Only Modulus
+
 ```k
     rule X modInt N => X
```
yzhang90: upstream

kwasm-lemmas.md
```diff
@@ -20,13 +20,198 @@ Basic logic
...
       requires 0 <=Int X
        andBool X  <Int N
       [simplification]
 ```
 
+`modInt` selects the least non-negative representative of a congruence class modulo `N`.
+
+```k
+    rule (X modInt M) modInt N => X modInt M
```
yzhang90: upstream

kwasm-lemmas.md
```diff
@@ -20,13 +20,198 @@ Basic logic
...
+  = r + y mod n
+  = (x mod m + y) mod n
+```
+
+#### Bit Shifting
+
+We want Z3 to understand what a bit-shift is.
+
+```k
+    rule (X >>Int N)          => 0 requires X <Int 2 ^Int N [simplification]
```
yzhang90: 135 - 136 upstream

kwasm-lemmas.md
```diff
@@ -20,13 +20,198 @@ Basic logic
...
+       andBool (2 ^Int N) modInt POW ==Int 0
+      [simplification]
+```
+
+Proof: These follow from the fact that shifting left by `n` bits is simply multiplying by `2^n`, and from previously proven rules of modular arithmetic.
+
+### Basic Operations
+
+```k
+    rule X  +Int 0 => X [simplification]
```
yzhang90: 207-212 upstream

Makefile
```diff
@@ -147,6 +147,9 @@ TEST_SYMBOLIC_BACKEND       := haskell
 
 tests/proofs/memory-concrete-type-spec.k%prove: PROVE_OPTIONS=--z3-tactic "(or-else (using-params smt :random-seed 1))" --z3-impl-timeout 5000
 tests/proofs/memory-concrete-type-spec.k%prove: KPROVE_MODULE=MEMORY-CONCRETE-TYPE-LEMMAS
+tests/proofs/wrc20-spec.k.prove: TEST_SYMBOLIC_BACKEND=java
+tests/proofs/wrc20-spec.k%prove: KPROVE_MODULE=WRC20-LEMMAS
+tests/proofs/locals-spec.k.prove: TEST_SYMBOLIC_BACKEND=java
```
ttuegel: Unless this proof has been modified, it should now pass on the Haskell backend:

```suggestion
tests/proofs/locals-spec.k.prove: TEST_SYMBOLIC_BACKEND=java
```

hjorthjort: I'm happy to try it, let's see what if it passes.

PR #261, opened 2019-11-18
Several updates to allow running proofs in Haskell.

traiansf:
* `kwasm-lemmas.md`: marked some simplification rules as such
* `tests/proofs/*-spec.k`: workaround a claims-detection frontend bug
* `Makefile`: updated to run all proofs with the Haskell backend


traiansf:
The `import` removals in this PR won't be necessary after https://github.com/kframework/k/pull/920 is merged and propagated; however I would like this to be merged, as it adds all proofs to continuous integration testing with the Haskell backend.

PR #277, opened 2019-12-25
Desugar imports before grouping the different kinds of module allocations.

### LINKED ISSUE #275: Module grouping messes up order of imports
Imports (for example function imports) can be defined either as `(func ... (import ...))` or `(import ... (func ...))`.
The former should just desugar to the latter, without changing the order of imports.
However, the way KWasm first groups the different definitions in a module, and then desugars them, the above imports would be reordered, changing their indices.

The following file works in the reference interpreter, but not in KWasm, where it results in `0`.

```
(module
  (func (export "foo") (result i32) (i32.const 0))
  (func (export "bar") (result i32) (i32.const 1))
)
(register "a")

(module
(func (import "a" "bar") (result i32))
(import "a" "foo" (func (result i32)))
(export "bar" (func 0) )
)

(assert_return (invoke "bar") (i32.const 1))
```

See branch
<https://github.com/kframework/wasm-semantics/tree/desugaring>

hjorthjort:
Addresses and closes #275.

### DIFF
--- a/tests/simple/desugaring.wast
+++ b/tests/simple/desugaring.wast
@@ -13,4 +13,14 @@
 
 (assert_return (invoke "bar") (i32.const 1))
 
+(module
+  (data (offset (i32.const 0)) "b")
+  (memory (data "a"))
+  (func (export "baz") (result i32)
+    (i32.load (i32.const 0))
+  )
+)
+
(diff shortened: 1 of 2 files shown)


hjorthjort:
I found further issues: we also need to deal with `data` and `elem` segments, which is less straightforward.

PR #279, opened 2019-12-30
Fix desugaring bug

### LINKED ISSUE #275: Module grouping messes up order of imports

### LINKED ISSUE #278: Change as many desugarings as possible in WASM-TEXT to be macros
This brings us closer to the official spec, and also avoids coordination mistakes such as #275.

hjorthjort:
Fixes #275, closes #278.

The specification (and reference interpreter supports this) says that all the abbreviations (syntactic sugars) in the text format are desugared immediately, and only after are the different definitions grouped. By not desugaring before grouping, we caused the order of imports, data declarations and elem declarations to get changed. This turns those desugarings into macros, so that they get applied before grouping.


hjorthjort:
This PR is stuck on this bug:

https://github.com/kframework/k/issues/1000

The issue is when there is a hex const after a `table` or `memory` inlined with `elem` or `data` such as the `"break-inner"` function in `block.wast`.

tests/simple/modules.wast
```diff
@@ -55,4 +55,24 @@
...
+    (local.set 0
+      (i32.add (local.get 0) (block (result i32) (i32.ctz (br 0 (i32.const 0x4)))))
+    )
+    (local.set 0
+      (i32.add (local.get 0) (block (result i32) (i32.ctz (block (result i32) (br 1 (i32.const 0x8))))))
+    )
+    (local.get 0)
+  )
+)
+
```
hjorthjort: Move this to the `desugaring.wast` test

tests/simple/desugaring.wast
```diff
@@ -0,0 +1,43 @@
...
+  (table funcref (elem 1))
+  (func (result i32) (i32.const 0))
+  (func (result i32) (i32.const 1))
+  (func (export "biz") (result i32)
+    (call_indirect (result i32) (i32.const 0))
+  )
+)
+
+(assert_return (invoke "biz") (i32.const 1))
+
```
hjorthjort:
```suggestion

;; Regression test: A module with hex integers inside a function after a `table` with inline `elem`.
(module
  (table funcref (elem))
  (func (export "break-inner") (result i32)
    (local i32)
    (local.set 0 (i32.const 0))
    (local.set 0 (i32.add (local.get 0) (block (result i32) (block (result i32) (br 1 (i32.const 0x1))))))
    (local.set 0 (i32.add (local.get 0) (block (result i32) (block (br 0)) (i32.const 0x2))))
    (local.set 0
      (i32.add (local.get 0) (block (result i32) (i32.ctz (br 0 (i32.const 0x4)))))
    )
    (local.set 0
      (i32.add (local.get 0) (block (result i32) (i32.ctz (block (result i32) (br 1 (i32.const 0x8))))))
    )
    (local.get 0)
  )
)

```

wasm-text.md
```diff
@@ -207,29 +230,46 @@ In the text format, it is also allowed to have a conditional without the `else`
...
 Intitial memory data, and initial table elements can be given inline in the text format.
+We supply macros that will unfold these definitions when they are part of a body of definitions, such as in a module.
+This is to ensure that the unfolding happens before the different elements in a module are grouped together, so as to maintain their order.
+
+We also supply rules for when the inlined definitions are encountered on top of the `<k>` cell, so that they can be desugared on the fly.
+
+This is useful when specifying modules in the more lax KWasm format, where they can be declared as they are needed.
 
 ```k
+    syntax Identifier ::= "#memId" | "#tabId"
```
ehildenb:
```suggestion
    syntax Identifier ::= "#memId" | "#tabId"
 // -----------------------------------------
```

An explanation of what these do would also be nice.

ehildenb: If they're just used when we need an anonymous table identifier or memory identifier, I would suggest `.Identifier`, or `.MemoryIdentifier` and `.TableIdentifier`.

hjorthjort: Good idea, thanks.

hjorthjort:
@ehildenb Can you have another look at this and see if we can get it merged?

ehildenb:
If you have to change a bunch to get this passing, please ask for re-review.

PR #280, opened 2020-01-08
Remove ocaml backend

ehildenb:


hjorthjort:
Did a search with `ag --ignore-case --hidden --ignore="deps/k" --ignore="tests/wasm-tests" "ocaml|opam" ; find . -name "*ocaml*"` and found the following things that may want to be addressed:

```
README.md
53:There are three backends of K available, the OCAML backend for concrete execution, the Java backend for symbolic reasoning and proofs, and the experimental Haskell backend for developers.
54:This repository generates the build-products for both backends in `.build/defn/java/` and `.build/defn/ocaml/`.
65:-   [Opam](https://opam.ocaml.org/doc/Install.html), **important**: Ubuntu users prior to 15.04 **must** build from source, as the Ubuntu install for 14.10 and prior is broken.
66:    `opam repository` also requires `rsync`.
78:    libyaml-dev lld-8 llvm llvm-8 llvm-8-tools maven opam openjdk-8-jdk pandoc        \
96:If you haven't already setup K's OCaml dependencies more recently than February 1, 2019, then you also need to setup the K OCaml dependencies:
99:./deps/k/k-distribution/src/main/scripts/bin/k-configure-opam-dev
116:To only build specific backends, you can do `make build-llvm`, `make build-java`, `make build-ocaml`, or `make build-haskell`.
```

I guess Opam is necessary to build K, but maybe we can ignore building the OCaml backend when building the dependency?
If nothing else, lines 53 and 116 need updating.

-----

```
Dockerfile
9:        libz3-dev lld-8 llvm-8-tools make maven netcat-openbsd opam          \
19:ADD deps/k/k-distribution/src/main/scripts/bin/k-configure-opam-dev deps/k/k-distribution/src/main/scripts/bin/k-configure-opam-common /home/user/.tmp-opam/bin/
20:ADD deps/k/k-distribution/src/main/scripts/lib/opam  /home/user/.tmp-opam/lib/opam/
22:    && ./.tmp-opam/bin/k-configure-opam-dev
```

If we can remove Opam dependencies, we can do so here as well.

------

```
tests/conformance/look_for_supported.sh
12:backends="llvm ocaml"
```

Update this line.

------

```
kwasm-lemmas.md
141:    // The Haskell and Java backend accept negative shifts (the LLVM and OCaml backends do not).
```

Update this line.

------

You should also remove these files:

```
./tests/conformance/unsupported-ocaml.txt
./tests/success-ocaml.out
```

ehildenb:
I believe I've addressed all the ones you pointed out, and re-ran your command to see if there were more, but couldn't find any.

PR #296, opened 2020-03-11
data, wasm: switch to using integer for take/drop lengths

ehildenb:
It's not immediately obvious to me that we can remove the `#take` and `#drop` altogether, so this just makes them use integer values instead.


ehildenb:
Screenshot from 2020-03-11 13-53-58

For example, here it says "Assert: due to *validation*, there at **at least** *n* values on the top of the stack."
1:55
Which to me means that we cannot assume there are **exactly** n values on the top of the stack.

PR #297, opened 2020-03-11
fix stmt infinite loop

sskeirik:
This branch fixes an infinite loop in stmt evaluation.

### DIFF
--- a/wasm.md
+++ b/wasm.md
@@ -147,9 +147,8 @@ The sorts `EmptyStmt` and `EmptyStmts` are administrative so that the empty list
     syntax Defns  ::= EmptyStmts
     syntax Stmts  ::= Instrs | Defns
  // --------------------------------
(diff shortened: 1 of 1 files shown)


ehildenb:
This was already fixed in #142 by adding the `:KItem`.

I would rather keep the existing fix because it keeps us from generating a bunch of extra empty `Stmts`, which means reducing the total number of steps of execution.

hjorthjort:
> This was already fixed in #294 by adding the `:KItem`.
> 
> I would rather keep the existing fix because it keeps us from generating a bunch of extra empty `Stmts`, which means reducing the total number of steps of execution.

@sskeirik Can you verify that this is correct? Does the proof for invoking a simple contract work on the latest master branch?

sskeirik:
Yes, the updated wasm-semantics already fixed this. Closing.

PR #303, opened 2020-03-23
Use the fast_balance verison of wrc20

hjorthjort:
Use the `fast_balance` version of the WRC20 (for Ewasm verification), and make the `#wrc20` macro a `ModuleDecl`.

This is so that we a) can use the `#wrc20` macro to set up a new contract in Ewasm (expects a Module, not any list of statements) and b) verify the fast balance version of the `do_balance` function, which is easier than the one which uses `i64.reverse_bytes`.


hjorthjort:
> I can't find the URL for the original fast_balance spec now. Why were additional calls to reverse_bytes added in the transfer function? I'm assuming that fast_balance means that transfer is slower?

It's the same URL as the fast_transfer, they are in the same place: https://gist.github.com/poemm/823566e89def007daa5d97ac5bd14419#file-wrc20_handwritten_fast_get_balance-wat

Yes, now transfer is slower, because bytes need to be reversed before doing arithmetic. The difference between the two contracts is basically if they store the little-endian, Wasm style integer balance for fast arithemetic (makes transfer faster), or if they store the big-endian, EVM style integer balance for being able to return it as is, without reversing (makes balance fast).

kwasm-lemmas.md
```diff
@@ -457,16 +457,16 @@ module WRC20
...
-    syntax TypeDecls ::= "#wrc20ReverseBytesTypeDecls"
- // --------------------------------------------------
-    rule #wrc20 => ( module #wrc20Body ) .EmptyStmts [macro]
+    syntax ModuleDecl ::= "#wrc20"
+    syntax Defns      ::= "#wrc20Body"
+    syntax Defns      ::= "#wrc20Imports"
+    syntax Defns      ::= "#wrc20Functions_fastBalance"
+    syntax Defns      ::= "#wrc20ReverseBytes"
+    syntax TypeDecls  ::= "#wrc20ReverseBytesTypeDecls"
+ // -------------------------------------
```
ehildenb:
```suggestion
    syntax TypeDecls  ::= "#wrc20ReverseBytesTypeDecls"
 // ---------------------------------------------------
```

PR #309, opened 2020-04-15
Allocator Statements

### LINKED ISSUE #308: Remove non-determinism due to implicit module creation
The following rules causes non-determinism:

```
    rule <k> D:Defn => ( module .Defns ) ~> D ... </k>
         <curModIdx> .Int </curModIdx>
```

The reason is that a definition can rewrite to another definition, for example

```
    rule <k> ( elem        OFFSET      ELEMSEGMENT ) =>     ( elem 0 OFFSET ELEMSEGMENT ) ... </k>
```

hjorthjort:
This separates the definitions of the functions, memories and tables (which can come in several different formats) from their actual allocations (which have specific requirements)

A few birds in one stone here:

1. Fixes https://github.com/kframework/ewasm-semantics/issues/15
2. Fixes #308 
3. Separates text-format definitions from their core-format behavior (of allocating `funcs`/`mems`/`globs`/`tables`. Now the definitions are purely a text format thing, and the allocations they trigger can't be directly expressed in a module.

It's still not quite the way it works in the official spec, where `alloc[func/def/mem/table]` is a function that acts on the entire store, but its closer.

I also removed cases where allocations cause `trap`. Those shouldn't really be trapping, it's a validity error.

### DIFF
--- a/wasm.md
+++ b/wasm.md
@@ -478,7 +478,7 @@ The importing and exporting parts of specifications are dealt with in the respec
     syntax Defn       ::= GlobalDefn
     syntax GlobalSpec ::= TextFormatGlobalType Instr
     syntax GlobalDefn ::= "(" "global" OptionalId GlobalSpec ")"
-    syntax Stmt       ::=     "global" OptionalId GlobalType
+                        |     "global" OptionalId GlobalType
  // --------------------------------------------------------
     rule <k> ( global OID:OptionalId TYP:TextFormatGlobalType IS:Instr ) => IS ~> global OID asGMut(TYP) ... </k>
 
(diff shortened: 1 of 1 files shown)


ehildenb:
I think a simpler fix would be to simply put an `[owise]` on the rule for module allocation:

```
rule <k> D:Defn => ( module .Defns ) ~> D ... </k>
     <curModIdx> .Int </curModIdx>
   [owise]
```

hjorthjort:
> I think a simpler fix would be to simply put an `[owise]` on the rule for module allocation:
> 
> ```
> rule <k> D:Defn => ( module .Defns ) ~> D ... </k>
>      <curModIdx> .Int </curModIdx>
>    [owise]
> ```

That's a good idea either way, to ensure we don't accidentally get non-determinism again. But other motivations for this change is to disallow the allocation statements inside module declarations (separation of text/core format) and to make the allocations consistent for globals, funcs, memories and tables.

ehildenb:
I don't think you need both changes (the owise and the subsort Alloc), do you?

hjorthjort:
Not necessary, but seems like a good failsafe for ending up with a similar situation as before in case intermediate allocators are added. It really should have been an `[owise]` to begin with, I think.

PR #312, opened 2020-04-16
kwasm-lemmas: only include this module when targeting symbolic backends

ehildenb:
This allows us to include the lemmas file into the rule merger for Polkadot, because otherwise the LLVM backend complains about function symbols on the LHS of rules.

### DIFF
--- a/kwasm-lemmas.md
+++ b/kwasm-lemmas.md
@@ -5,7 +5,7 @@ These lemmas aid in verifying WebAssembly programs behavior.
 They are part of the *trusted* base, and so should be scrutinized carefully.
 
 ```k
-module KWASM-LEMMAS
+module KWASM-LEMMAS [symbolic]
     imports WASM-TEXT
     imports WRC20
 ```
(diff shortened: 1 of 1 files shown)


hjorthjort:
I thought we wanted to keep all lemmas in one file rather than spread out. Maybe it would at least be good to name all the lemmas files `*-lemmas.md`?

ehildenb:
Well lemmas that are specific for wrc20 should be in a `wrc20-lemmas.k` next to the `wrc20-spec.k` file, or just in the `wrc20-spec.k` file. Lemmas that are re-usable between many proofs and situations should be factored out into `kwasm-lemmas.md` (which is a large body), but ones specifically for `wrc20` should be in its file.

If there are some lemmas in wrc20-lemmas that are general, they should be moved to other parts of `kwasm-lemmas.md`.

ehildenb:
I've separated out the WRC20 separation steps into this PR: #314.

PR #314, opened 2020-04-19
Separate WRC20 from KWasm Lemmas

ehildenb:
After #310.


wrc20.md
```diff
@@ -0,0 +1,314 @@
...
+endmodule
+```
+
+Macros
+------
+
+The following module gives macros for the wrc20 code so that its parts can be expressed succinctly in proofs.
+
+```k
+module WRC20
```
hjorthjort, line 77: The definitoin of the WRC20 is in this module. That used to be imported to KWASM-LEMMAS, but isn't anymore (good choice).
So either import both WRC20 and WRC20-LEMMAS in the WRC20-SPEC, or merge the WRC20 and the WRC20-LEMMAS modules.

PR #320, opened 2020-04-21
Remove java backend

### LINKED ISSUE #229: Example proof using KWasm demo
In repository kframework/wasm-semantics

First make sure you can do `make deps` and `make build-java`.

Have to make sure KLab is on your PATH: instructions in KLab readme: https://github.com/dapphub/klab

you need to patch KLab (make sure to use same commit I'm using `3eb35c4`:

```
--- a/lib/behavior.js
+++ b/lib/behavior.js
@@ -87,11 +87,11 @@ const genBehaviour = (state, init = null) => {
       let k = state.nodes[node.id.split("_")[0]];
       if(k) {
         evm = kast.get(k, "ethereum.evm");
-        if(evm) {
-            node.revert = kast.get(k, "ethereum.evm.statusCode").label == "EVMC_REVERT_NETWORK";
-            node.oog = kast.get(k, "ethereum.evm.statusCode").label == "EVMC_OUT_OF_GAS_NETWORK";
-            node.gas = kast.get(k, "ethereum.evm.callState.gas");
-        }
+        //if(evm) {
+        //    node.revert = kast.get(k, "ethereum.evm.statusCode").label == "EVMC_REVERT_NETWORK";
+        //    node.oog = kast.get(k, "ethereum.evm.statusCode").label == "EVMC_OUT_OF_GAS_NETWORK";
+        //    node.gas = kast.get(k, "ethereum.evm.callState.gas");
+        //}
       }
       if(circular[node]) node.status = "circular";
       if(node.revert) {
```

Set the environment variable KLAB_OUT (I have it mapped to a ramdisk for speed, on my machine it's about 12% improvement).

Comment out all the proofs except the one you want, and then run:

```
./kwasm klab-prove tests/proofs/simple-arithmetic-spec.k --def-module KWASM-LEMMAS
```

It will output some JSON blob, which should probably contain `#True` in it somewhere (if the proof succeeds).

Then you want to visualize the proof:

```
./kwasm klab-view tests/proofs/simple-arithmetic-spec.k
```

Instructions for navigating the proof are in KLab readme.

Basics are `Shift-n` and `Shift-p` for "next" and "previous" state. The views I turn on are `t` (current term), `c` (current constraint), and `r` (last rule applied).

ehildenb:
Blocked on #319 (will be easier to review once #319 is in).

This removes the Java backend completely from the repository, so that we can cleanly switch to the Bytes sort.

Currently nothing in the repo depends on Java backend.

-----

Closes #229


Makefile
```diff
@@ -108,29 +99,21 @@ $(haskell_kompiled): $(haskell_defn)
...
+TEST_CONCRETE_BACKEND := llvm
+TEST_SYMBOLIC_BACKEND := haskell
 
 KPROVE_MODULE := KWASM-LEMMAS
 
-PROVE_OPTIONS ?=
+KPROVE_OPTIONS ?=
+
+tests/proofs/memory-concrete-type-spec.k%prove: KPROVE_MODULE = MEMORY-CONCRETE-TYPE-LEMMAS
+tests/proofs/wrc20-spec.k%prove:                KPROVE_MODULE = WRC20-LEMMAS
```
hjorthjort, line 116: You can remove the `%` here I think, since there is no longer a `klab-prove` target.

ehildenb: Done

PR #321, opened 2020-04-21
Remove ByteMap sort, use Bytes directly

ehildenb:
Blocked on #320.

This removes the ByteMap sort in favor of using Bytes directly.

All the proofs pass, but some are significantly slower. I had to disable some lemmas which are no longer true.


ehildenb:
Broken into several smaller PRs.

PR #327, opened 2020-04-23
Use macros for block instructions

hjorthjort:
This helps when we do preprocessing, because it gives us fewer cases to deal with in the text format when we do simplifications.


hjorthjort:
Subsumed by the new `text2abstract` function from #333, which is wehere we should do such desugaring operations.

PR #330, opened 2020-04-24
Remove context lookup for locals

hjorthjort:
Here's a first draft of using "higher order functions" to traverse the tree of modules. It is not the most elegant solution. But it's a first step.


ehildenb:
Subsumed by #333

PR #331, opened 2020-05-02
Use Dockerhub images for CI

ehildenb:
This uses the published Dockerhub images for K on CI instead of building it in the Jenkinsfile, which will make the overall CI time less.

Both the `Makefile` and the `kwasm` script are modified to have the following behavior about where to pull `K_RELEASE` from (and thus `krun`, `kprove`, `kore-exec`, etc...):

-   If the user sets `K_RELEASE`, use that version.
-   If the submodule is built, use the submodule version.
-   Otherwise, use `which kompile` to find a global or user install of K, and use that build.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,9 +6,13 @@ DEPS_DIR  := deps
 DEFN_DIR  := $(BUILD_DIR)/defn
 
 K_SUBMODULE := $(DEPS_DIR)/k
-K_RELEASE   ?= $(K_SUBMODULE)/k-distribution/target/release/k
-K_BIN       := $(K_RELEASE)/bin
-K_LIB       := $(K_RELEASE)/lib
+ifneq (,$(shell which kompile))
+  K_RELEASE ?= $(dir $(shell which kompile))..
+else
+  K_RELEASE ?= $(K_SUBMODULE)/k-distribution/target/release/k
+endif
+K_BIN := $(K_RELEASE)/bin
+K_LIB := $(K_RELEASE)/lib
 
 K_BUILD_TYPE := Debug
 
--- a/Makefile
+++ b/Makefile
@@ -6,10 +6,10 @@ DEPS_DIR  := deps
 DEFN_DIR  := $(BUILD_DIR)/defn
 
 K_SUBMODULE := $(DEPS_DIR)/k
-ifneq (,$(shell which kompile))
-  K_RELEASE ?= $(dir $(shell which kompile))..
-else
+ifneq (,$(wildcard deps/k/k-distribution/target/release/k/bin/*))
   K_RELEASE ?= $(K_SUBMODULE)/k-distribution/target/release/k
+else
+  K_RELEASE ?= $(dir $(shell which kompile))..
 endif
 K_BIN := $(K_RELEASE)/bin
 K_LIB := $(K_RELEASE)/lib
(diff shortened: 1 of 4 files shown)


ehildenb:
Not ready for review yet actually.

Makefile
```diff
@@ -88,13 +88,13 @@ build-llvm:    $(llvm_kompiled)
 build-haskell: $(haskell_kompiled)
 
 $(llvm_kompiled): $(llvm_defn)
-	$(K_BIN)/kompile --backend llvm                                           \
+	kompile --backend llvm                                                    \
```
hjorthjort, line 91: Is this relevant to this PR, or just cleanup? I see that in the beginning of the Makefile we are exporting a new path with the `K_BIN` on top anyway, so I guess it isn't necessary at all.

ehildenb: We need this, because in the Docker image K is not installed at `deps/k/...`, but at `/usr/bin`.

hjorthjort: So is the `export PATH` in the Makefile a potential problem? If we were to build the K dependency (even if we don't do that now I guess that could change later) that would introduce a new `kompile` binary in the path and we would silently switch to using that one.

It seems the "correct" way to to this would be to have the Jenkins-file set the `K_BIN` variable when calling `make build`.

ehildenb: Well, the Makefile should work in all these scenarios:

-   User sets `K_RELEASE`, and so manually picks which version of K to select from (this is for the Haskell backend team, which often uses out-of-tree builds of K).
-   User doesn't set `K_RELEASE`, but builds the submodule, we should prefer the submodule build of K then.
-    User doesn't set `K_RELEASE`, and doesn't build the submodule, but has a global install of K available, we should then use the global install of K.

ehildenb: This isn't quite what happens, but I can correct this PR (and others) once the K PR gets in.

Makefile
```diff
@@ -6,9 +6,14 @@ DEPS_DIR  := deps
...
-K_RELEASE   ?= $(K_SUBMODULE)/k-distribution/target/release/k
-K_BIN       := $(K_RELEASE)/bin
-K_LIB       := $(K_RELEASE)/lib
+ifneq (,$(wildcard deps/k/k-distribution/target/release/k/bin/*))
+  K_RELEASE ?= $(K_SUBMODULE)/k-distribution/target/release/k
+else
+  K_RELEASE ?= $(dir $(shell which kompile))..
+endif
+K_BIN := $(K_RELEASE)/bin
+K_LIB := $(K_RELEASE)/lib
```
dwightguth, line 15: pretty sure this should be lib/kframework

Makefile
```diff
@@ -6,9 +6,14 @@ DEPS_DIR  := deps
 DEFN_DIR  := $(BUILD_DIR)/defn
 
 K_SUBMODULE := $(DEPS_DIR)/k
-K_RELEASE   ?= $(K_SUBMODULE)/k-distribution/target/release/k
-K_BIN       := $(K_RELEASE)/bin
-K_LIB       := $(K_RELEASE)/lib
+ifneq (,$(wildcard deps/k/k-distribution/target/release/k/bin/*))
+  K_RELEASE ?= $(K_SUBMODULE)/k-distribution/target/release/k
+else
+  K_RELEASE ?= $(dir $(shell which kompile))..
```
dwightguth, line 12: `/..`

PR #333, opened 2020-05-11
Get rid of #ContextLookup with a preprocessing function

hjorthjort:
This removes `#ContextLookup` for locals. It does so by creating a preprocessing function with some context. Currently, the context is only the local ids, but we can add the other kind of identifiers, label depth, and all other text-format specific stuff to the context and get rid of it all in a single pass.


wasm.md
```diff
@@ -1303,6 +1289,7 @@ Exports make functions, tables, memories and globals available for importing int
 ```k
     syntax Defn       ::= ExportDefn
     syntax ExportDefn ::= "(" "export" WasmString "(" Externval ")" ")"
+    syntax Alloc      ::= ExportDefn
```
hjorthjort, line 1293: Formatting

wasm.md
```diff
@@ -1327,6 +1314,7 @@ The value of a global gets copied when it is imported.
                         | "(" "table"  OptionalId TableType            ")" [klabel( tabImportDesc)]
                         | "(" "memory" OptionalId MemType              ")" [klabel( memImportDesc)]
                         | "(" "global" OptionalId TextFormatGlobalType ")" [klabel(globImportDesc)]
+    syntax Alloc      ::= ImportDefn
```
hjorthjort, line 1317: Formatting

wasm.md
```diff
@@ -1496,7 +1484,6 @@ Then, the surrounding `module` tag is discarded, and the definitions are execute
 ```k
     syntax Stmt       ::= ModuleDecl
     syntax ModuleDecl ::= "(" "module" OptionalId Defns ")"
-                        |     "module" OptionalId Map
```
hjorthjort, line 1499: Formatting

wasm-text.md
```diff
@@ -360,6 +360,99 @@ Imports can be declared like regular functions, memories, etc., by giving an inl
     rule ( memory OID:OptionalId (import MOD NAME) MT:MemType   ) => ( import MOD NAME (memory OID MT  ) )  [macro]
 ```
 
+Desugaring
+----------
```
hjorthjort, line 364: Structure this section more, give some markdown explanations.

test.md
```diff
@@ -348,7 +348,6 @@ This simply checks that the given function exists in the `<funcs>` cell and has
            <funcAddrs> ... #ContextLookup(IDS , TFIDX) |-> FADDR ... </funcAddrs>
            ...
          </moduleInst>
-         <nextFuncAddr> NEXT </nextFuncAddr>
```
hjorthjort, line 351: Random clean-up: this cell wasn't used in the rule

wasm-text.md
```diff
@@ -360,6 +360,229 @@ Imports can be declared like regular functions, memories, etc., by giving an inl
...
+* Remove module names.
+* Give the text format and core format different types, and have the preprocessing handle the conversion. So that identifiers don't even exist in the core type.
+
+### The Context
+
+The `Context` contains information of how to map text-level identifiers to corresponding indices.
+
+```k
+    syntax Context    ::= ctx(localIds: Map)
+ // ----------------------------------------
```
ehildenb, line 386: Formatting

```suggestion
    syntax Context ::= ctx(localIds: Map)
 // -------------------------------------
```

wasm-text.md
```diff
@@ -360,6 +360,229 @@ Imports can be declared like regular functions, memories, etc., by giving an inl
...
+```
+
+### Instructions
+
+TODO: Desugar folded instructions.
+
+```k
+    syntax Instr ::= "#t2aInstr" "<" Context ">" "(" Instr ")" [function]
+ // ---------------------------------------------------------------------
+
```
ehildenb, line 431:
```suggestion
```

wasm-text.md
```diff
@@ -360,6 +360,229 @@ Imports can be declared like regular functions, memories, etc., by giving an inl
...
+    rule #t2aInstr<_>(call F)           => call F
+    rule #t2aInstr<_>(call_indirect TU) => call_indirect TU
+```
+
+#### Parametric Instructions
+
+```k
+    rule #t2aInstr<_>(drop)   => drop
+    rule #t2aInstr<_>(select) => select
+```
```
ehildenb, line 454: I assume we're not doing this with an `[owise]` rule because of our discussion about halting if translation fails?

I'm still on the fence about it, putting in an owise rule would make it easier to extend the translation.

Another option would be to combine all these rules with `#Or` patterns, but IMO it's a bit uglier (eg. `rule #t2aInstr<_>((drop #Or select) #as I) => I`)

wasm-text.md
```diff
@@ -360,6 +360,229 @@ Imports can be declared like regular functions, memories, etc., by giving an inl
...
+    rule #t2aStmt<C>(I:Instr) => #t2aInstr<C>(I)
+    rule #t2aStmt<_>(S) => S [owise]
+
+    rule #t2aDefn<C>(( func OID:OptionalId FS:FuncSpec )) => ( func OID #t2aFuncSpec<C>(FS))
+    rule #t2aDefn<C>(D:Defn) => D [owise]
+
+    rule #t2aFuncSpec<C> (( export WS ) FS:FuncSpec ) => ( export WS ) #t2aFuncSpec<C>(FS)
+    rule #t2aFuncSpec<C>(T:TypeUse LS:LocalDecls IS:Instrs) => #t2aFuncSpec<C>(#updating(T LS IS))
+    rule #t2aFuncSpec<ctx(... localIds: (_ => #ids2Idxs(T, LS)))>(#updating(T:TypeUse LS:LocalDecls IS:Instrs) => #ready(T LS IS))
+    rule #t2aFuncSpec<C>(#ready(T:TypeUse LS:LocalDecls IS:Instrs)) => #t2aTypeUse<C>(T) #t2aLocalDecls<C>(LS) #t2aInstrs<C>(IS)
```
ehildenb, line 414: This whole `#updating` and `#ready` thing is confusing. Does it not work with something like:

```
rule #t2aFuncSpec<C>(T:TypeUse LS:LocalDecls IS:Instrs) => #t2aTypeUse<#ids2Idxs(T, LS)> #t2aLocalDecls<#ids2Idxs(T, LS)>(LS) #t2aInstrs<#ids2Idxs(T, LS)>(IS)
```

wasm-text.md
```diff
@@ -360,6 +360,229 @@ Imports can be declared like regular functions, memories, etc., by giving an inl
...
+```
+
+### Functions for Gathering Context
+
+The following are helper functions for gathering and updating context.
+
+```k
+    syntax Map        ::= #ids2Idxs(TypeUse, LocalDecls)      [function, functional]
+                        | #ids2Idxs(Int, TypeUse, LocalDecls) [function, functional]
+ // --------------------------------------------------------------------------------
```
ehildenb, line 566:
```suggestion
    syntax Map ::= #ids2Idxs(TypeUse, LocalDecls)      [function, functional]
                 | #ids2Idxs(Int, TypeUse, LocalDecls) [function, functional]
 // -------------------------------------------------------------------------
```

wasm-text.md
```diff
@@ -360,6 +360,229 @@ Imports can be declared like regular functions, memories, etc., by giving an inl
...
+      => (ID |-> N) #ids2Idxs(N +Int 1, TDS, LDS)
+    rule #ids2Idxs(N, (param _) TDS, LDS)   => #ids2Idxs(N +Int 1, TDS, LDS)
+    rule #ids2Idxs(N, TD:TypeDecl TDS, LDS) => #ids2Idxs(N       , TDS, LDS) [owise]
+
+    rule #ids2Idxs(N, .TypeDecls, local ID:Identifier _ LDS:LocalDecls)
+      => (ID |-> N) #ids2Idxs(N +Int 1, .TypeDecls, LDS)
+    rule #ids2Idxs(N, .TypeDecls, LD:LocalDecl LDS) => #ids2Idxs(N +Int 1, .TypeDecls, LDS) [owise]
+
+    syntax FuncSpec   ::= "#updating" "(" FuncSpec ")" | "#ready" "(" FuncSpec ")"
+ // ------------------------------------------------------------------------------
```
ehildenb, line 583: I don't think these should be needed, and either way they should be moved closer to where they are used (and the formatting fixed).

wasm-text.md
```diff
@@ -360,6 +360,229 @@ Imports can be declared like regular functions, memories, etc., by giving an inl
...
+                        | #ids2Idxs(Int, TypeUse, LocalDecls) [function, functional]
+ // --------------------------------------------------------------------------------
+    rule #ids2Idxs(TU, LDS) => #ids2Idxs(0, TU, LDS)
+
+    rule #ids2Idxs(_, .TypeDecls, .LocalDecls) => .Map
+    rule #ids2Idxs(N, (type _)    , LDS) => #ids2Idxs(N, .TypeDecls, LDS)
+    rule #ids2Idxs(N, (type _) TDS, LDS) => #ids2Idxs(N, TDS       , LDS)
+
+    rule #ids2Idxs(N, (param ID:Identifier _) TDS, LDS)
+      => (ID |-> N) #ids2Idxs(N +Int 1, TDS, LDS)
```
ehildenb, line 574: You might prefer the syntax `#ids2Idxs(N +Int 1, TDS, LDS) [ ID <- N ]`, which means something different than what you have written.

What you have written will halt with a K runtime error if the same `ID` is entered into the map twice (and so we're attempting a map union over two maps with the same key). The example I wrote will continue executing, but will prefer later definitions for the same identifier.

Not sure if it's a valid thing to have the same identifier show up multiple times, but something for you to consider.

wasm-text.md
```diff
@@ -360,6 +360,229 @@ Imports can be declared like regular functions, memories, etc., by giving an inl
...
+    rule #ids2Idxs(N, (param _) TDS, LDS)   => #ids2Idxs(N +Int 1, TDS, LDS)
+    rule #ids2Idxs(N, TD:TypeDecl TDS, LDS) => #ids2Idxs(N       , TDS, LDS) [owise]
```
ehildenb, line 576:
```suggestion
    rule #ids2Idxs(N, (param _)   TDS, LDS) => #ids2Idxs(N +Int 1, TDS, LDS)
    rule #ids2Idxs(N, TD:TypeDecl TDS, LDS) => #ids2Idxs(N       , TDS, LDS) [owise]
```

wasm-text.md
hjorthjort: That would work now, but the context is a record because it's supposed to contain more maps. We would want to only mention relevant parts of that context when updating it. I could use your suggestion for now, but it would have to be changed as soon as we expand the context.

The `#fun(ctx(...))(C)` doesn't work either due to a parsing issue that isn't solved yet.

How about making a context-updating function with a relevant name? Like
```
syntax Context ::= #updateLocalIds(Context, Map) [function, functional] | #updateLocalIds(Context, Map, Status) [function, functional]
syntax Status   ::= "#updating" | "#done"

rule #updateLocalIds(C, M) => #updateLocalids(C, M, #updating)
rule #updateLocalIds(ctx(... localIds: (_ => M)), M, #updating => #ready)
rule #updateLocalIds(C, _, #ready) => C
```

```
rule #t2aFuncSpec<C>(T:TypeUse LS:LocalDecls IS:Instrs)
  => #t2aTypeUse   <#updateLocalIds(C, #ids2Idxs(T, LS))>(T)
     #t2aLocalDecls<#updateLocalIds(C, #ids2Idxs(T, LS))>(LS)
     #t2aInstrs    <#updateLocalIds(C, #ids2Idxs(T, LS))>(IS)
```

It's not pretty, but at least it hides the complexity under a relevant name.

hjorthjort: Or we could just skip using the record and use a map from `String` to `Map`. In the end, the context will contain at least 7 maps (locals, labels, functions, types, globals, memories, tables) and one integer (label depth) if we want to get rid of `#ContextLookup`, so always mentioning the whole context would become ugly fast.

wasm-text.md
hjorthjort: Yes, that, and for similarity with the official spec. It just enumerates every concrete instruction and translates it to abstract syntax.

We could also keep the `[owise]` and exit abnormally somehow when we encounter an invalid reference. Like

```
syntax {Sort} Sort ::= #validationError(WasmString)

rule #t2aInstr<ctx(... localIds: LIDS)(local.get ID:Identifier) => #validationError("Local " +String id2String(ID) +String " not declared in function body")
```

wasm-text.md
hjorthjort: The same identifier can't appear twice.

> An identifier context is well-formed if no index space contains duplicate identifiers.
https://webassembly.github.io/spec/core/text/conventions.html#contexts

So runtime error on transforming to abstract syntax seems more appropriate than overwriting. Again, we could exit abnormally with some message.

wasm-text.md
ehildenb: The context-updating function seems like the best choice right now to me, because it allows us to use the nice record type.

Just define it at the same place the `Context` record is defined:

```
syntax Context ::= ctx ( localIds: Map )
            | #updateLocalIds ( Context , Map ) [function]
            | #updateLocalIdsAux ( Context , Map , Status ) [function]
// ------------------------
// Definition of `#updateLocalIds` and `#updateLocalIdsAux` here.
```

Until we have record-update syntax, this will have to do.

wasm-text.md
ehildenb: If the official spec enumerates all possibilities, let's also do that, because then it's easier to audit.

ehildenb: Can be resolved.

hjorthjort: To be clear, the official spec enumerates literally _every_ instruction separately:

```
i32.load m
i64.load m
f32.load m
f64.load m
i32.load8_s m
i32.load8_u m
i32.load16_s m
i32.load16_u m
i64.load8_s m
i64.load8_u m
i64.load16_s m
i64.load16_u m
i64.load32_s m
i64.load32_u m
i32.store m
i64.store m
f32.store m
f64.store m
i32.store8 m
i32.store16 m
i64.store8 m
i64.store16 m
i64.store32 m
memory.size
memory.grow

Comment (hjorthjort) [reply wasm-text.md:453]: I collapsed these a bit. I'm all for full explicitness, but right now all the `#t2a` functions take and return the same sort, and in the abstract syntax, things are more grouped. An end goal with `text2abstract` is a clean separation, to the degree that the `#t2a` functions take a `TextX` sort and returns an `X` sort. When it comes to that, full enumeration in the style of the spec makes sense.

File: wasm-text.md [hunk 14]
```diff
@@ -360,6 +360,233 @@ Imports can be declared like regular functions, memories, etc., by giving an inl
...
+Some classes of invalid programs, such as those where an identifier appears in a context in which it is not declared, are undefined.
+
+The function deals with the desugarings which are context dependent.
+Other desugarings are either left for runtime or expressed as macros (for now).
+
+TODO:
+
+* Get rid of inline type declarations (only allow types defined first, inline type declarations serve as documentation and identifier bindings). Something like `(func (type X) TDS:TDecls ... ) => (func (type X))` and `(func TDS:TDecls ...) => (type TDECLS) (func (type NEXT_TYPE_ID) or something)`
+* Remove module names.
+* Give the text format and core format different types, and have the preprocessing handle the conversion. So that identifiers don't even exist in the core type.
```
ehildenb, line 378: The first bullet point here is a bit confusing, can you be more clear?

wasm.md
```diff
@@ -1012,12 +1004,12 @@ Sort `Signedness` is defined in module `BYTES`.
 
 ```k
     syntax Instr ::= "load" "{" IValType Int Int Signedness"}"
- // ----------------------------------------------------------
+                   | IValType "." LoadOp Int
```
ehildenb, line 1007: Why is this turned into an `Instr` instead of a `PlainInstr`?

hjorthjort: Because `PlainInstr` is only for instructions that may appear in source text. This is an administrative instruction. It may be better for format it as such, or maybe use a function instead.

wasm-text.md
hjorthjort: Sure, how about this:

```suggestion
* Get rid of inline type declarations. The text format allows specifying the type directly in the function header using the `param` and `result` keywords. However, these will either be desugared to a new top-level `type` declaration or they must match an existing one. In the abstract format, a function's type is a pointer to a top-level `type` declaration. This could either be an immediate desugaring, like `(func (type X) TDS:TDecls ... ) => (func (type X))` and `(func TDS:TDecls ...) => (type TDECLS) (func (type NEXT_TYPE_ID)`, or an initial pass could gather all the function types and declare them)
* Remove module names.
* Give the text format and core format different types, and have the preprocessing handle the conversion. So that identifiers don't even exist in the core type.
```

ehildenb: Looks better for sure. My only nit is that I prefer `-   ` instead of `* ` (so dash followed by three spaces) for bullet lists. And I prefer one sentence per line.

PR #336, opened 2020-05-18
Make wrap operate over bytes

ehildenb:
Some small cleanup that came out of #321.

Every time we use `#wrap`, it's operating over bytes, so this changes the first argument of wrap to be a byte-width instead of a bit-width, which makes using it in the context of `#setRange` and `#getRange` simpler.

Many of the `#wrap` lemmas will go away in the next PR, but for now it would be good to just audit all the uses of `#wrap` in the semantics on this branch and make sure that they are correct for the new semantics.

I also combine the two memory proofs into a single proof file.


tests/proofs/functions-spec.k
```diff
@@ -0,0 +1,17 @@
...
+module FUNCTIONS-SPEC
+    imports FUNCTIONS-LEMMAS
+
+    rule <k> #wrap(2, #getRange(ByteMap <| BM0 |>, ADDR, 4)) => #getRange(ByteMap <| BM0 |>, ADDR, 2) ... </k>
+      requires #isByteMap(ByteMap <| BM0 |>)
+       andBool 0 <=Int ADDR
+
+    rule <k> #setRange ( ByteMap <| BM0 |> , ADDR , #wrap ( 2 , #getRange ( ByteMap <| BM0 |> , ADDR , 4 ) ) , 2 ) => ByteMap <| BM0 |> ... </k>
+      requires #isByteMap(ByteMap <| BM0 |>)
+       andBool 0 <=Int ADDR
```
hjorthjort, line 16: You don't need to expose the constructor here, just let the map be symbolic. Then you also don't need to change this when we switch to `Bytes`.

Also, one proof used constructor-like formatting (`#getRange ( A , B , C )`) and another function-like formatting (`#getRange(A, B, C)`). I switched to the latter here, but I'm okay with either.

```suggestion
    rule <k> #wrap(2, #getRange(BM, ADDR, 4)) => #getRange(BM, ADDR, 2) ... </k>
      requires #isByteMap(BM)
       andBool 0 <=Int ADDR

    rule <k> #setRange(BM, ADDR, #wrap(2, #getRange(BM, ADDR, 4)), 2) => BM ... </k>
      requires #isByteMap(BM)
       andBool 0 <=Int ADDR
```

ehildenb: Fixed.

PR #339, opened 2020-05-20
kwasm-lemmas: remove unused module

ehildenb:


hjorthjort:
I think there was talk about upstreaming these into K. Is that still interesting?

PR #340, opened 2020-05-20
Cleanups and organization

ehildenb:
-   Several lemmas over `#get` removed.
-   Re-organize and minor documentation of `kwasm-lemmas.md`.
-   Standardization of comparison directions (eg. `0 <Int WIDTH` instead of `WIDTH >Int 0`)

This may be easier to review commit-by-commit with `git log -p`, and in some cases `git show COMMIT_ID -b`

### DIFF
--- a/kwasm-lemmas.md
+++ b/kwasm-lemmas.md
@@ -354,17 +354,6 @@ Lemmas about `#getRange` and `#setRange` simplification:
        andBool WIDTH1 *Int 8 ==Int SHIFT
       [simplification]
 
-    rule #getRange(#setRange(BM, EA, VALUE, SET_WIDTH), EA, GET_WIDTH)
-      => #wrap(GET_WIDTH, VALUE)
-      requires         GET_WIDTH <=Int SET_WIDTH
-      [simplification]
-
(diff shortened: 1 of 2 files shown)


kwasm-lemmas.md
```diff
@@ -321,52 +321,39 @@ Lemmas about `#getRange` and `#setRange` simplification:
...
-      => #wrap(GET_WIDTH, VALUE)
-      requires         GET_WIDTH <=Int SET_WIDTH
+       andBool 8 <=Int SHIFT
+       andBool #isByteMap(BM)
       [simplification]
 
-    rule #getRange(#setRange(BM, EA, VALUE, SET_WIDTH), EA, GET_WIDTH)
-      => #wrap(SET_WIDTH, VALUE)
-      requires (notBool GET_WIDTH <=Int SET_WIDTH)
-       andBool #getRange(BM, EA +Int SET_WIDTH, GET_WIDTH -Int SET_WIDTH) ==Int 0
```
hjorthjort, line 365: I take it these `#getRange(#setRange(...))` lemmas are no longer needed? Were any other removed?

hjorthjort: (As opposed to just moved)

PR #345, opened 2020-05-26
Unfold definitions

hjorthjort:
In preparation for the extraction of identifiers and removing `#ContextLookup` for definitions of tables, functions, globals, and memories.

This passes each module through an `unfoldDefns` function. Previously we were doing some of this unfolding at parse time, with macros, and some at runtime. Now we do it all ahead of time.

Some tests are updated to reflect that we no longer support unfolding at runtime, so only definitions that are declared properly in modules can be desugared. We could in principle support all the syntactic sugars, but I think it would make the desugaring code a lot messier and verbose (and departing from the official spec), since we would have to deal with definitions interleaved with statements while keeping track of which module we are in.

This one might be easiest to review commit-by-commit.

### DIFF
--- a/wasm-text.md
+++ b/wasm-text.md
@@ -254,19 +254,6 @@ Note that it is possible to define multiple exports inline, i.e. export a single
           ...
          </k>
 
-    syntax FuncSpec   ::= InlineExport FuncSpec
- // -------------------------------------------
-    rule <k> ( func                         EXPO:InlineExport SPEC:FuncSpec )
-          => ( func #freshIdRuntime(NEXTID) EXPO              SPEC          )
-          ...
-         </k>
-         <nextFreshId> NEXTID => NEXTID +Int 1 </nextFreshId>
-
-    rule <k> ( func ID:Identifier ( export ENAME ) SPEC:FuncSpec )
-          => ( export ENAME ( func ID ) )
-          ~> ( func ID SPEC )
-          ...
-         </k>
 ```
 
 ### Imports
@@ -278,10 +265,8 @@ Imports can be declared like regular functions, memories, etc., by giving an inl
  // --------------------------------------------------------------
 
     syntax GlobalSpec ::= InlineImport TextFormatGlobalType
-    syntax FuncSpec   ::= InlineImport TypeUse
  // --------------------------------------------
     rule ( global OID:OptionalId (import MOD NAME) TYP          ) => ( import MOD NAME (global OID TYP ) )  [macro]
-    rule ( func   OID:OptionalId (import MOD NAME) TUSE         ) => ( import MOD NAME (func   OID TUSE) )  [macro]
 ```
 
 Desugaring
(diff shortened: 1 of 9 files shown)


wasm-text.md
```diff
@@ -229,161 +229,155 @@ In the text format, it is also allowed to have a conditional without the `else`
...
-          ~> ( elem  ID (offset (i32.const 0) .Instrs) ES )
-          ...
-         </k>
+```k
+    syntax Defns ::= unfoldDefns  ( Defns )       [function]
+                   | #unfoldDefns ( Defns , Int ) [function]
+ // --------------------------------------------------------
+    rule unfoldDefns(DS) => #unfoldDefns(DS, 0)
+    rule #unfoldDefns(.Defns, _) => .Defns
+    rule #unfoldDefns(D:Defn DS, I) => D #unfoldDefns(DS, I) [owise]
```
ehildenb, line 272:
```suggestion
    rule unfoldDefns(DS) => #unfoldDefns(DS, 0)
    
    rule #unfoldDefns(.Defns   , _) => .Defns
    rule #unfoldDefns(D:Defn DS, I) => D #unfoldDefns(DS, I) [owise]
```

wasm-text.md
```diff
@@ -411,9 +405,10 @@ Since we do not have polymorphic functions available, we define one function per
     syntax TypeUse   ::= "#t2aTypeUse"    "<" Context ">" "(" TypeUse    ")" [function]
     syntax LocalDecl ::= "#t2aLocalDecl"  "<" Context ">" "(" LocalDecl  ")" [function]
  // -----------------------------------------------------------------------------------
-    rule text2abstract(SS) => #t2aStmts<ctx( ... localIds: .Map)>(SS)
+    rule text2abstract(DS:Defns) => text2abstract(( module DS ) .Stmts)
+    rule text2abstract(SS)       => #t2aStmts<ctx( ... localIds: .Map)>(SS) [owise]
 
-    rule #t2aStmt<C>(( module OID:OptionalId DS )) => ( module OID #t2aDefns<C>(DS) )
+    rule #t2aStmt<C>(( module OID:OptionalId DS )) => ( module OID #t2aDefns<C>(unfoldDefns(DS)) )
```
ehildenb, line 411: I'm not sure it makes sense to entangle these two functions does it? Is it possible to do the `unfold` pass completely then the `#t2a` pass? Then `#text2abstract(D) => #t2a<...>(unfoldDefns(...)` or something like that?

ehildenb: Not sure it makes sense, just thinking out loud here.

hjorthjort: We could do an initial pass that only does unfolding. So we could do

```
unfoldStmts(( module OID:OptionalId DS) SS) => (module OID unfoldDefns(DS)) unfoldStmts(SS)
```

Then maybe a module structuring pass:

```
structureModules((module OID:OptionalId DS) SS) => sortModule((module OID DS)) structureModules(SS)
```

And then finally the existing `#t2a` pass (we can have a better name maybe, now that it is more specialized.

```
#t2aStmts<_>(#module(... funcs: FS, tables: TS, .......) #as M)
=> #t2aModule<#freshModuleContext(FS, TS, ...)>(M)
```
where `#freshModuleContext` produces a context with all the identifers extracted.

Then we get
```
text2abstract(DS:Defns) => text2abstract((module DS) .Stmts)
text2abstract(SS) => #t2aStmts(structureModules((unfoldStmts(SS))) [owise]
```

hjorthjort: `#t2aModule` would just distrubte the `#t2aDefns` with the new context out over all the different defintion types in the record.

hjorthjort: Done

PR #346, opened 2020-05-27
Sort text module

hjorthjort:
In preparation for the extraction of identifiers and removing #ContextLookup for definitions of tables, functions, globals, and memories.

Comes after #345.

This makes the module sorting part of the `text2abstract` transformation. This is both in line with how the official specification is written and will make extracting the correct identifiers cleaner and faster.

### DIFF
--- a/test.md
+++ b/test.md
@@ -19,6 +19,18 @@ module WASM-TEST
     imports WASM-TEXT
 ```
 
+Bare Allocations
+----------------
+
+We allow allocations to appear outside of modules, for example interleaved with assertions in tests.
+This is purely a KWasm feature, which is useful for testing.
+
+```k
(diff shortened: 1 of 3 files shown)


ehildenb:
@hjorthjort it looks like what happened is that because the pr #345 was merged, and this one was made into that branch, this one was closed when that one was merged. I think if you want to avoid this in the future, better to just open stacked PRs against `master` but mark them as Draft until they are ready for merged.

hjorthjort:
Yeah, this was just so I would have something to point to to explain where
#345 was headed. I'll push up a new one

Everett Hildenbrandt <notifications@github.com> schrieb am Sa., 30. Mai
2020, 9:24 AM:

hjorthjort:
Reopened here, but still not done: https://github.com/kframework/wasm-semantics/pull/349

PR #349, opened 2020-05-30
Sort text module

hjorthjort:
This PR moves us closer to the official spec by:
1. Structuring modules fully into separate lists for each kind of definition.
2. Performing the module structuring as part of the transformation from text format to abstract format.

### DIFF
--- a/wasm-text.md
+++ b/wasm-text.md
@@ -471,7 +471,18 @@ Since we do not have polymorphic functions available, we define one function per
     rule #t2aStmt<C>(I:Instr) => #t2aInstr<C>(I)
     rule #t2aStmt<_>(S) => S [owise]
 
-    rule #t2aModuleDecl<C>(#module(... funcs: FS) #as M) => #updateFuncs(M, #t2aDefns<C>(FS))
+    rule #t2aModuleDecl<C>(#module(... id: OID, types: TS, funcs: FS, tables: TABS, mems: MS, globals: GS, elem: EL, data: DAT, start: S,  importDefns: IS, exports: ES))
+      => #module(... id: OID,
+                     types: TS,
+                     funcs: #t2aDefns<C>(FS),
+                     tables: TABS,
+                     mems: MS,
+                     globals: GS,
+                     elem: EL,
+                     data: DAT,
+                     start: S,
+                     importDefns: IS,
+                     exports: ES)
 
     rule #t2aDefn<C>(( func OID:OptionalId FS:FuncSpec )) => ( func OID #t2aFuncSpec<C>(FS))
     rule #t2aDefn<C>(D:Defn) => D [owise]
@@ -487,17 +498,6 @@ Since we do not have polymorphic functions available, we define one function per
 
     rule #t2aLocalDecl<C>(local ID:Identifier AVT:AValType) => local AVT .ValTypes
     rule #t2aLocalDecl<C>(LD) => LD [owise]
-```
-
-Since we can't update record as we are evaluating a function, we use helper functions for this.
-
-```k
-    syntax ModuleDecl ::= #updateFuncs    ( ModuleDecl, Defns )       [function]
-                        | #updateFuncsAux ( ModuleDecl, Defns, Bool ) [function]
- // ----------------------------------------------------------------------------
-    rule #updateFuncs(M, FS) => #updateFuncsAux(M, FS, false)
-    rule #updateFuncsAux(#module(... funcs: _ => FS), FS, false => true)
-    rule #updateFuncsAux(M, _ , true) => M
 ```
 
 ### Instructions
--- a/wasm-text.md
+++ b/wasm-text.md
@@ -389,27 +389,27 @@ In doing so, the respective ordering of all types of definitions are preserved.
 ```k
     syntax Stmts ::= structureModules ( Stmts ) [function]
  // ------------------------------------------------------
-    rule structureModules((module OID:OptionalId DS) SS) => sortModule((module OID DS)) structureModules(SS)
+    rule structureModules((module OID:OptionalId DS) SS) => structureModule((module OID DS)) structureModules(SS)
     rule structureModules(.Stmts) => .Stmts
     rule structureModules(S SS) => S strucutreModules(SS) [owise]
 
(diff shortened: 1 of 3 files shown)


wasm-text.md
```diff
@@ -369,8 +379,50 @@ Other desugarings are either left for runtime or expressed as macros (for now).
...
+    rule #structureModule((T:TypeDefn   DS:Defns => DS), #module(... types: (TS => T TS)))
+    rule #structureModule((I:ImportDefn DS:Defns => DS), #module(... importDefns: (IS => I IS)))
+    rule #structureModule((X:FuncDefn   DS:Defns => DS), #module(... funcs: (FS => X FS)))
+    rule #structureModule((X:GlobalDefn DS:Defns => DS), #module(... globals: (GS => X GS)))
+    rule #structureModule((T:TableDefn  DS:Defns => DS), #module(... tables: (TS => T TS)))
+    rule #structureModule((M:MemoryDefn DS:Defns => DS), #module(... mems: (MS => M MS)))
+    rule #structureModule((E:ExportDefn DS:Defns => DS), #module(... exports: (ES => E ES)))
+    rule #structureModule((I:DataDefn   DS:Defns => DS), #module(... data: (IS => I IS)))
+    rule #structureModule((I:ElemDefn   DS:Defns => DS), #module(... elem: (IS => I IS)))
+    rule #structureModule((S:StartDefn  DS:Defns => DS), #module(... start: (_ => S .Defns)))
```
ehildenb, line 412:
```suggestion
    rule #structureModule((T:TypeDefn   DS:Defns => DS), #module(... types:       TS => T TS))
    rule #structureModule((I:ImportDefn DS:Defns => DS), #module(... importDefns: IS => I IS))
    rule #structureModule((X:FuncDefn   DS:Defns => DS), #module(... funcs:       FS => X FS))
    rule #structureModule((X:GlobalDefn DS:Defns => DS), #module(... globals:     GS => X GS))
    rule #structureModule((T:TableDefn  DS:Defns => DS), #module(... tables:      TS => T TS))
    rule #structureModule((M:MemoryDefn DS:Defns => DS), #module(... mems:        MS => M MS))
    rule #structureModule((E:ExportDefn DS:Defns => DS), #module(... exports:     ES => E ES))
    rule #structureModule((I:DataDefn   DS:Defns => DS), #module(... data:        IS => I IS))
    rule #structureModule((I:ElemDefn   DS:Defns => DS), #module(... elem:        IS => I IS))
    rule #structureModule((S:StartDefn  DS:Defns => DS), #module(... start:       _  => S .Defns))
```

wasm-text.md
```diff
@@ -398,36 +450,48 @@ Record updates can currently not be done in a function rule which also does othe
...
+    syntax Stmt       ::= "#t2aStmt"       "<" Context ">" "(" Stmt       ")" [function]
+    syntax ModuleDecl ::= "#t2aModuleDecl" "<" Context ">" "(" ModuleDecl ")" [function]
+    syntax Defn       ::= "#t2aDefn"       "<" Context ">" "(" Defn       ")" [function]
+    syntax FuncSpec   ::= "#t2aFuncSpec"   "<" Context ">" "(" FuncSpec   ")" [function]
+    syntax TypeUse    ::= "#t2aTypeUse"    "<" Context ">" "(" TypeUse    ")" [function]
+    syntax LocalDecl  ::= "#t2aLocalDecl"  "<" Context ">" "(" LocalDecl  ")" [function]
+ // ------------------------------------------------------------------------------------
     rule text2abstract(DS:Defns) => text2abstract(( module DS ) .Stmts)
-    rule text2abstract(SS)       => #t2aStmts<ctx( ... localIds: .Map)>(unfoldStmts(SS)) [owise]
+    rule text2abstract(SS)       => #t2aStmts<ctx( ... localIds: .Map)>(structureModules(unfoldStmts(SS))) [owise]
```
ehildenb, line 467: So here we're mixing `structureModules` directly into the `#t2aStmts` pass.

Does it make sense to instead run `structureModules` across the whole definition, then run `#t2a*`? (similar to how we're running `unfoldDefinitions` before running `#t2a*`?

hjorthjort: Well, it's still one at a time, right? The conversion `text2abstract` first unfolds, then structures, then does `#t2a`. Like function chaingin: `#unfold ; structure ; #t2a`.

ehildenb: Ah, I'm sorry, I was reading it incorrectly, I thought the lhs was #t2a.

hjorthjort:
DeepinScreenshot_select-area_20200605131401
The text format modules look like this. It's the same for all the definition types really, they are just records. Here's functions:
DeepinScreenshot_select-area_20200605131606
In my ideal world, we would not use the parenthesis style at all in the core format, but a more record-like structure.
Using K records make sense for modules, since they have so many many parts, but definitions have so few fields we might as well use a simple strucutre, like `syntax FuncDefn ::= "#func" TypeIdx Valtypes Instrs` for example, or `syntax FuncDefn ::= "#func" "{" TypeIdx "," Valtypes "," Instrs "}"`.
We could also commit to using the parenthesis style in the core format, but we should try to make it unambiguous so that execution of the core format is as simple as possible. And I think it would look nicer if we didn't use the parenthesis style  in the core format then, to ensure there is a clear visible distinction of core format and text format.

hjorthjort:
I'll have a look later today on the formatting changes.

ehildenb:
I like the change to using records, but I think we should be consistent. I don't see any disadvantage to using records for everything.

If you have:

```
syntax Record ::= record ( a: A , b: B , c: C )
```

You can still instantiate it as:

```
rule <k> foo => record(aaa, bbb, ccc) ... </k>
```

without using the record syntax. The record syntax is just a different way to instantiate and pattern match on it.

hjorthjort:
Great, even better.

ehildenb:
But to be clear, the change to record syntax can happen in another PR so that we get the #ContextLookup stuff unblocked.

hjorthjort:
Yes, I was thinking that after the `#ContextLookup` thing is done, every type of definition should be passing through a `#t2a` case, and then we can update them to record form. After that, we can even give the text and abstract format different sorts, which would be nice. But it's not a priority.

PR #355, opened 2020-06-05
Remove contextlookup for functions

hjorthjort:
The tests are updated, because we now don't support function IDs outside of modules, and the ID isn't stored, so the assertions can't check for them.

### DIFF
--- a/wasm.md
+++ b/wasm.md
@@ -33,7 +33,6 @@ Configuration
             <typeIds>     .Map </typeIds>
             <types>       .Map </types>
             <nextTypeIdx> 0    </nextTypeIdx>
-            <funcIds>     .Map </funcIds>
             <funcAddrs>   .Map </funcAddrs>
             <nextFuncIdx> 0    </nextFuncIdx>
             <tabIds>      .Map </tabIds>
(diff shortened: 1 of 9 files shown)


wasm.md
```diff
@@ -764,12 +762,11 @@ The `#take` function will return the parameter stack in the reversed order, then
...
     syntax PlainInstr ::= "call" Index
  // ----------------------------------
-    rule <k> call TFIDX => ( invoke FADDR:Int ) ... </k>
+    rule <k> call IDX:Int => ( invoke FADDR:Int ) ... </k>
          <curModIdx> CUR </curModIdx>
          <moduleInst>
            <modIdx> CUR </modIdx>
-           <funcIds> IDS </funcIds>
-           <funcAddrs> ... #ContextLookup(IDS , TFIDX) |-> FADDR ... </funcAddrs>
+           <funcAddrs> ... IDX |-> FADDR ... </funcAddrs>
```
ehildenb, line 769: I think it makes more sense to put the `FADDR:Int` in the `<funcAddrs>` cell, not the `<k>` cell, because it shows up on the LHS in the `<funcAddrs>` cell and the rHS in the `<k>` cell.

ehildenb: (the sort check, that is)

But also, by putting it in an `(invoke FADDR)`, you're asserting that it's the same sort that `(invoke_)` takes, which is either `Index` or `Int` (not sure), so that alone may be enough.

hjorthjort: `invoke` takes `Int` so the sort check is unnecessary

PR #361, opened 2020-06-14
Update .PHONY target

hjorthjort:
See finding explained here: https://github.com/kframework/ewasm-semantics/pull/26#discussion_r439817967

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,7 @@ PATH := $(K_BIN):$(PATH)
 export PATH
 
 .PHONY: all clean deps                                                     \
-        build build-llvm build-haskell                                     \
+        build build-%                                                      \
         test test-execution test-simple test-prove                         \
         test-prove-good test-prove-bad                                     \
         test-conformance test-conformance-parse test-conformance-supported \


hjorthjort:
I don't really see why `make` wouldn't resolve match the `build-llvm` and `build-haskell` targets against the `build-%` target.
I'm not sure, however, that adding the `build-%` target achieves anything. The dependency checks in the build of the ewasm-semantics work just as well when `build-%` isn't declared `.PHONY` at all.

ehildenb:
@hjorthjort this change doesn't make sense here, `build-llvm` and `build-haskell` are explicitly declared in the `Makefile` here.

I don't think adding `build-%` is doing what you think in the ewasm semantics. The `%` in make is not a wildcard, it's a pattern match. So if you don't have a `%` on the LHS of the `:`, it is being populated with nothing. So I'm pretty sure you're just adding `build-` to the `.PHONY` there.

In your ewasm PR you can have:

```
.PHONY: build

build: build-llvm build-haskell

build-%: ...
```

So that we can do `make build` to get the most common builds, have `build` be `.PHONY`, and have the one rule for `build-%`.

Though, honestly, I'm inclined to say you should just have explicit `build-llvm` and `build-haskell` targets in the ewasm makefile. Your choice though.


Anyway, this one can be closed I think.

PR #367, opened 2020-07-06
Type desugaring during text -> core phase

hjorthjort:
This is in preparation of getting rid of `#ContextLookup` for types.

### DIFF
--- a/wasm-text.md
+++ b/wasm-text.md
@@ -297,9 +297,25 @@ Other desugarings are either left for runtime or expressed as macros (for now).
     rule #unfoldDefns(D:Defn DS, I, M) => D #unfoldDefns(DS, I, M) [owise]
 ```
 
-#### Functions
+#### Types
 
-TODO: Unfold type-use into type declarations.
+A type can be be declared implicitly when a `TypeUse` is a `TypeDecls`.
+In this case it will allocate a type when the type is not in the current module instance.
+
+```k
+    rule #unfoldDefns(( func OID:OptionalId TDECLS:TypeDecls LOCALS:LocalDecls BODY:Instrs ) DS, I, M)
+      => (func OID (type {M [asFuncType(TDECLS)]}:>Identifier) TDECLS LOCALS BODY)
+         #unfoldDefns(DS, I, M)
+      requires         asFuncType(TDECLS) in_keys(M)
+
+    rule #unfoldDefns(( func OID:OptionalId TDECLS:TypeDecls LOCALS:LocalDecls BODY:Instrs ) DS, I, M)
+      => (type #freshId(I) (func TDECLS))
+         (func OID (type #freshId(I)) TDECLS LOCALS BODY)
+         #unfoldDefns(DS, I +Int 1, M [asFuncType(TDECLS) <- #freshId(I)])
+      requires notBool asFuncType(TDECLS) in_keys(M)
+```
+
+#### Functions
 
 ```k
     syntax FuncDefn ::= "(" "func" OptionalId  FuncSpec ")"
--- a/wasm-text.md
+++ b/wasm-text.md
@@ -303,6 +303,18 @@ A type can be be declared implicitly when a `TypeUse` is a `TypeDecls`.
 In this case it will allocate a type when the type is not in the current module instance.
 
 ```k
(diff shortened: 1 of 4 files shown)


wasm-text.md
```diff
@@ -229,6 +229,15 @@ In the text format, it is also allowed to have a conditional without the `else`
        andBool ( ID ==K OID'' orBool notBool isIdentifier(OID'') )
 ```
 
+### Types
+
+```k
+    syntax TypeDefn ::= "(type" OptionalId "(" "func" TypeDecls ")" ")"
+ // -------------------------------------------------------------------
+
```
ehildenb, line 237:
```suggestion
```

wasm-text.md
```diff
@@ -275,17 +284,47 @@ Other desugarings are either left for runtime or expressed as macros (for now).
...
+    rule #unfoldDefns(( func OID:OptionalId TDECLS:TypeDecls LOCALS:LocalDecls BODY:Instrs ) DS, I, M)
+      => (func OID (type {M [asFuncType(TDECLS)]}:>Identifier) TDECLS LOCALS BODY)
+         #unfoldDefns(DS, I, M)
+      requires         asFuncType(TDECLS) in_keys(M)
+
+    rule #unfoldDefns(( func OID:OptionalId TDECLS:TypeDecls LOCALS:LocalDecls BODY:Instrs ) DS, I, M)
+      => (type #freshId(I) (func TDECLS))
+         (func OID (type #freshId(I)) TDECLS LOCALS BODY)
+         #unfoldDefns(DS, I +Int 1, M [asFuncType(TDECLS) <- #freshId(I)])
+      requires notBool asFuncType(TDECLS) in_keys(M)
```
ehildenb, line 327: Why do we use `#freshId` here instead of just taking `I`? I thought the purpose of `I` is that it would be a fresh identifier not scene anywhere yet?

ehildenb:
Looks pretty good, one question about the unfolding procedure for types.

wasm-text.md
hjorthjort: It's an `Int`, not an `Identifier`. Using an `Int` to type the function is allowed (selecting the `X`th type of the module as the type for the function). But the actual indices we give the things in a module are strict. The first `type` to appear must get index 0, the next one index 1, etc. The production for type declarations is `(type OID (func TDECLS)`. It can take an identifier, but we can't assign it an index.

We could skip using identifiers, but then we would have to keep track of exactly which index the type will get, but that would require some extra bookkeeping. The same type can be declared more than once (but can't be generated implicitly more than once), so we can't just look at the size of the map.

hjorthjort:
@ehildenb I added some complexity to ensure newly declared types end up last in the modul, but it removed the use of `freshId`. This also takes us closer to the official spec, I'd say, which says new implicitly declared types should simply be placed last in the module.

wasm-text.md
hjorthjort: In the end I did add that extra bookkeeping, see c1d0558

wasm-text.md
```diff
@@ -275,17 +283,58 @@ Other desugarings are either left for runtime or expressed as macros (for now).
...
+ // ------------------------------------------------------------------
+    rule types2indices(DS) => #types2indices(DS, #ti(... t2i: .Map, count: 0))
+
+    rule #types2indices(.Defns, TI) => TI
+
+    rule #types2indices((type OID (func TDECLS)) DS, #ti(... t2i: M, count: N)) => #types2indices(DS, #ti(... t2i: M [ asFuncType(TDECLS) <- N ], count: N +Int 1))
+      requires notBool asFuncType(TDECLS) in_keys(M)
+
+    rule #types2indices((type OID (func TDECLS)) DS, #ti(... t2i: M, count: N)) => #types2indices(DS, #ti(... t2i: M                            , count: N +Int 1))
+      requires         asFuncType(TDECLS) in_keys(M)
```
ehildenb, line 335: You could probably fold these two rules together into one by removing the side conditions and using `orDefault` from `domains.k`, if you prefer:

```
M [ asFuncType(TDECLS) <- (M [ asFuncType(TDECLS) ] orDefault N) ]
```

PR #376, opened 2020-08-28
KWasm updates pulled over from KEVM

ehildenb:
This PR does 4 things:

-   Adds the `--bug-report` option directly to `kwasm` runner for ease of access.
-   Uses the now frontend-exposed `--debugger --debug-script` for the `--repl` option.
-   Makes the increased stack/memory limits for `krun/kprove` globally set.
-   Updates the argument/option parsing to be the same as KEVM.

### DIFF
--- a/kwasm
+++ b/kwasm
@@ -68,7 +68,7 @@ run_prove() {
     def_module="$1" ; shift
 
     additional_proof_args=()
-    ! $repl       || additional_proof_args+=(--debugger --debug-script kast.kscript)
+    ! $repl       || additional_proof_args+=(--debugger)
     ! $bug_report || additional_proof_args+=(--haskell-backend-command "kore-exec --bug-report $bug_report_name")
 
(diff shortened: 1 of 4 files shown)


kwasm
```diff
@@ -59,22 +63,25 @@ run_kast() {
...
-    if $repl; then
-        haskell_backend_command="$k_release_dir/bin/kore-repl --repl-script $kwasm_dir/kast.kscript"
-    fi
-    kprove --directory "$backend_dir" "$run_file" "$@" --haskell-backend-command "$haskell_backend_command"
+    local additional_proof_args
+
+    def_module="$1" ; shift
+
+    additional_proof_args=()
+    ! $repl       || additional_proof_args+=(--debugger --debug-script kast.kscript)
```
hjorthjort, line 71: Can you remove the `--debug-script`, that file isn't available anymore.

hjorthjort (changes requested):

kwasm
ehildenb: Done.

hjorthjort (approved):

PR #383, opened 2020-11-05
Binary parser for instructions

hjorthjort:
This contains the meat of the binary parser.

It spearates text and abstract format where possible.


sskeirik (approved):
May want to build the wasm parser separately (as opposed to wasm-text) so that we can be sure that the binary-to-kast conversion only uses the wasm abstract syntax.

This will just encourage clean separation between the two layers.

PR #386, opened 2020-11-17
Binary parser finish

hjorthjort:


sskeirik (approved):
Approved, though one comment that I think will be helpful is noting in the binary parser readme or perhaps in the `kwasm_ast.py` source code as a header comment that the `kwasm_ast.py` source file depends crucially on the klabels for the WASM abstract syntax productions and must be kept in sync for them to work correctly.

PR #387, opened 2020-11-26
Bugfix/workaround for creating bytes lists for Kast

hjorthjort:
The string representation of bytes that Kast needs is hard to achieve in Python.
This is a workaround to that.
It may be more brittle than necessary, so suggestions are welcome.


hjorthjort:
@ehildenb @sskeirik Any ideas how we can solve this problem better?

andrei-marinica (approved):

ehildenb:
This looks OK to me.

ehildenb (approved):

PR #392, opened 2020-12-18
Filename metadata

hjorthjort:
Store the filename of the module as an optional metadata field

### DIFF
--- a/binary-parser/kwasm_ast.py
+++ b/binary-parser/kwasm_ast.py
@@ -53,10 +53,6 @@ EXPORT = 'aExportDefn'
 DEFNS  = '___WASM-COMMON-SYNTAX_Defns_Defn_Defns'
 INSTRS = '___WASM-COMMON-SYNTAX_Instrs_Instr_Instrs'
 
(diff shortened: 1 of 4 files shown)


sskeirik (approved):
These changes look good to me.

As an aside, since I'm out of the loop on the day-to-day status of the project, for my own understanding, I would like to know more about the context of PRs, i.e., what larger goal are you aiming towards here --- better tooling, spec compliance, etc...?

hjorthjort:
Food feedback, thank you. I know you've said so before and I will try to keep that in mind more in the future.

The context is this:
I'm working on producing coverage metrics for Wasm code. That is on a per-file basis: deciding whether a given test has touched a particular part of a particular source file. Since Wasm modules are read from a file and then loaded into the state, without any source reference, I need some way to determine what file produced a given module. A single file can also produce serveras different module instances, so it's not enough to look at the per-module instance coverage: the same source code can be loaded several times into several module instances.

This is essentially an upstreaming of changes necessary for the elrond-semantics repo, which is driving this change.

The changes in the PR are those required to keep track of the mapping ( file -> module instance ), plus some relevant refactoring encountered along the way.

PR #414, opened 2021-04-05
introduce BlockMetaData for measureing wasm coverage

yzhang90:
We introduce
```
    syntax BlockMetaData ::= #funcBlockInfo ( Int ) [klabel(funcBlockInfo), symbol]
                                            | #blockInfo     ( Int ) [klabel(blockInfo), symbol]
```
In wasm_kast.py, we can only assign a unique #blockInfo(block_id) to #block, #if and #loop, not the function.

the `#funcBlockInfo` is used in the following rule
```
    rule <instrs> ( invoke FADDR )
               => init_locals #revs(#take(lengthValTypes(TDOMAIN), VALSTACK)) ++ #zero(TLOCALS)
               ~> #block([TRANGE], INSTRS, #funcBlockInfo(FADDR))
               ~> frame MODIDX TRANGE #drop(lengthValTypes(TDOMAIN), VALSTACK) LOCAL
               ...
         </instrs>
```

### DIFF
--- a/binary-parser/kwasm_ast.py
+++ b/binary-parser/kwasm_ast.py
@@ -154,14 +154,14 @@ def global_type(mut, valtype):
 NOP = KApply('aNop', [])
 UNREACHABLE = KApply('aUnreachable', [])
 
-def BLOCK(vec_type, instrs):
-    return KApply('aBlock', [vec_type, instrs])
+def BLOCK(vec_type, instrs, block_info):
+    return KApply('aBlock', [vec_type, instrs, block_info])
 
-def IF(vec_type, then_instrs, else_instrs):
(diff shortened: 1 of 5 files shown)


hjorthjort (changes requested):
Looks good, but I think we should skip the `#funcBlockInfo`. We already give the function metadata, and the block created when invoking a function doesn't really "exist", we just insert it for simplicity. In the real semantics, they insert a label instead.

I think we can just use `-1` for that block.

Also, it's probably better if we use `OptionalInt` instead of `Int`, then we can use `.Int` for "ghost" blocks instead of `-1`. That also makes it easier to ignore those blocks in side conditions.

hjorthjort (approved):

PR #437, opened 2021-09-29
Add some simple proofs

calin1304:
Added some claims I came up with as an exercise
1. For any X, Y equal integers, X - Y = 0.
2. For any X, Y integers, maximum of X and Y is greater than or equal to X and to Y
3. The sum of even integers is even.


kwasm-lemmas.md
```diff
@@ -21,6 +21,11 @@ Basic logic
 Basic arithmetic
 ----------------
 
+```k
+    rule #signed(_, 0) => 0 [simplification]
+    rule #signed(_, X -Int Y) => X -Int Y [simplification]
```
ehildenb, line 26: This isn't true. Consider the definition of `#signed` (in `data.md`)

```
    rule #signed(ITYPE, N) => N                  requires 0            <=Int N andBool N <Int #pow1(ITYPE) 
    rule #signed(ITYPE, N) => N -Int #pow(ITYPE) requires #pow1(ITYPE) <=Int N andBool N <Int #pow (ITYPE) 
```

So if we have that `notBool (0 <=Int X -Int Y andBool X - Int Y <Int #pow1(ITYPE))`, and we also have that `#pow1(ITYPE) <=Int X -Int Y andBool X -Int Y <Int #pow(ITYPE)`, we should be returning `(X -Int Y) -Int #pow(ITYPE)`, but your simplification rule returns `X -Int Y`.

ehildenb: You can add a requires clause to your second rule which says `0 <=Int X -Int Y andBool X -Int Y <Int #pow1(ITYPE)`, but that would be saying the same thing as the original rule in the semantics.

So instead, we should make it so the prover can know that `0 <=Int X -Int Y andBool X -Int Y <Int #pow1(ITYPE)` is true. Which proof is this lemma needed for?

tests/proofs/simple-spec.k
```diff
@@ -0,0 +1,56 @@
...
+            ITYPE:IValType . const Y ~>
+            ITYPE:IValType . ge_u    ~>
+            #if([ITYPE:IValType .ValTypes], ITYPE:IValType.const X, ITYPE:IValType.const Y, _)
+            => .
+            ...
+        </instrs>
+        <valstack> S => < ITYPE > ?MAX : S </valstack>
+    requires
+        #inUnsignedRange(ITYPE, X) andBool
+        #inUnsignedRange(ITYPE, Y)
```
ehildenb, line 31:
```suggestion
    requires #inUnsignedRange(ITYPE, X)
     andBool #inUnsignedRange(ITYPE, Y)
```

tests/proofs/simple-spec.k
```diff
@@ -0,0 +1,56 @@
+requires "kwasm-lemmas.md"
```
ehildenb, line 1: For `#signed(ITYPE, 0)`, it probably can't tell that `0 <Int #pow1(ITYPE)`, even though `ITYPE` can only be `i32` or `i64`.

```
rule 0 <Int #pow1(_) => true [simplification]
```

and see if this works on Ana's branch.

Make sure to check the defniition of `#pow1` and make sure this is true.

hjorthjort (approved):

PR #459, opened 2023-03-16
Fix `KBytes`

bbyalcinkaya:
This PR fixes the `KBytes` function in `binary-parser`, which converts Python `bytes` to K `Bytes`. The current implementation puts extra backslashes.  

```python
>>> from kwasm_ast import KBytes
>>> KBytes(bytes([0, 65, 255])).token
'b"\\x00A\\xff"'
>>> KBytes(bytes([0, 65, 255])).token == 'b"\x00\x41\xff"'
False
>>> KBytes(bytes([0x0, 0x41, 0xff])).token == 'b"\x00\x41\xff"'
False
```

After the fix:
```python
>>> from kwasm_ast import KBytes
>>> KBytes(bytes([0, 65, 255])).token == 'b"\x00A\xff"'
True
>>> KBytes(bytes([0, 65, 255])).token == 'b"\x00\x41\xff"'
True
>>> KBytes(bytes([0x0, 0x41, 0xff])).token == 'b"\x00\x41\xff"'
True
```

This solution is taken from [`avm-semantics`](https://github.com/runtimeverification/avm-semantics/blob/35a75917cad492bd37485db82345d218f9c4bd74/kavm/src/kavm/pyk_utils.py#L181).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -213,7 +213,10 @@ tests/%.wasm.bparse: tests/%.wasm pykwasm-poetry-install
 
 binary_parser_tests:=$(wildcard tests/binary/*.wat)
 
-test-binary-parser: $(binary_parser_tests:.wat=.wasm.bparse)
+test-binary-parser: $(binary_parser_tests:.wat=.wasm.bparse) test-kwasm-ast
+
+test-kwasm-ast: pykwasm-poetry-install
+	poetry -C pykwasm run pytest binary-parser/test_kwasm_ast.py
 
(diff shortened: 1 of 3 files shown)


hjorthjort (changes requested):
Please also add a test case that exercises this change

hjorthjort:
I'm glad to see this is fixed! As I recall, this was a problem with the old frontend.

hjorthjort (approved):

PR #553, opened 2023-12-14
Implement table and reference instructions

bbyalcinkaya:


wasm-text.md
```diff
@@ -1045,6 +1225,74 @@ The following are helper functions for gathering and updating context.
     rule #idcFuncsAux(.Defns, (func      _:FuncSpec) FS => FS, IDX => IDX +Int 1, _ACC)
     rule #idcFuncsAux(.Defns, .Defns, _, ACC) => ACC
 
+    syntax Map ::= #idcTables    (Defns, Defns ) [function]
```
virgil-serbanuta, line 1228: Aren't these functions total?

The same question for the similar functions below.

virgil-serbanuta (commented):
This looks good, I'll take another look after you answer my comments.

wasm-text.md
bbyalcinkaya: Some functions in this module are non-total because when doing text-to-abstract conversion they perform some validations. For example `#idcTables` checks that the table IDs are unique. If there are duplicate IDs, the result is undefined.

virgil-serbanuta:
Please notify me when this is ready for re-review.

virgil-serbanuta (approved):

PR #604, opened 2024-04-08
Set up Nix flake build

Baltoli:
This will enable us to allow `kup install kwasm`, and will make it easier for us to package the `mx-*` family of tools that depend on the WASM semantics with Nix / kup in the future.

The code is adapted from a mix of other semantics' Flake builds, but in the process of doing so I might have done something unusual / suboptimal, so please review critically!


goodlyrottenapple (approved):
looks good!

Baltoli:
Superseded by: https://github.com/runtimeverification/wasm-semantics/pull/622

PR #612, opened 2024-04-17
Make memory a constructor-based list

virgil-serbanuta:


pykwasm/src/pykwasm/kdist/wasm-semantics/data/sparse-bytes.k
```diff
@@ -6,22 +6,11 @@ module SPARSE-BYTES
...
-          ]
-  syntax SparseBytes ::= ".SparseBytes"
-          [ function, total, hook(LIST.unit), klabel(.SparseBytes),
-            symbol, smtlib(smt_seq_nil)
-          ]
-  syntax SparseBytes ::= SBChunk(SBItem)
-          [ function, total, hook(LIST.element), klabel(SparseBytesItem),
-            symbol, smtlib(smt_seq_elem)
-          ]
+  syntax SparseBytes ::= List{SBItemChunk, ""}
```
bbyalcinkaya, line 13: We can also use `SBItem` here directly but it may require too many changes in downstream semantics.

bbyalcinkaya (approved):
Looks good to me. Please see my comment below.

virgil-serbanuta:
@bbyalcinkaya I increased the SMT timeout for one test, we can discuss that if needed, and I can fix or rollback if you think it wasn't the right thing to do.

PR #618, opened 2024-04-24
Remove `kwasm.sh`

tothtamas28:
Removes `kwasm.sh`, and implements commands `kwasm {run,kast,prove}` in `kwasm.py` instead.


ehildenb:
@tothtamas28 please start including descriptions in teh PR main text. It only takes a minute to write.

virgil-serbanuta (approved):

PR #622, opened 2024-04-30
Set up Nix flake build

Baltoli:
This PR sets up a Nix flake build and associated CI workflows for the WASM semantics that will allow us to do similar for the MX- projects, and to get KWASM from `kup`. It is a revised version of #604 that accommodates @tothtamas28's recent changes to the Python packaging of the semantics; broadly speaking the changes are very similar and largely relate to different paths and environment variables being necessary for the Python-packaged version.


goodlyrottenapple (approved):
looks good

PR #631, opened 2024-05-14
Make `numberCvtOp` total

bbyalcinkaya:
This PR aims to prevent Booster's fallback on conversion operations, includes the following changes:

* Define `IWidth` to represent integer widths and avoid using `Int` whenever possible. 
* Remove `#signedWidth` function.
* Make `numberCvtOp` total.


jberthold (approved):
Looks good. 
I will still test using a local set-up what this does for the fall-backs but it is definitely an improvement, approving :+1:

PR #642, opened 2024-06-10
Remove #distinctBits.

virgil-serbanuta:
The following rule matched way too many comparisons, which were transformed by the `requires` clause into two parallel comparisons and a `#distinctBits(VAL1, VAL2)` term that would never evaluate to true. The better solution is to match the terms in the main rule, not in `#distinctBits`, even if that makes the rule a bit uglier.
```
rule VAL1 +Int VAL2 <Int MAX => true requires VAL1 <Int MAX andBool VAL2 <Int MAX andBool #distinctBits(VAL1, VAL2) [simplification]
```

Preliminary tests show that this PR likely fixes the main Coindrip performance problems.


jberthold (approved):
I tested this PR with the `mx-backend` proofs, running with booster (It was with a prior version of the PR, before the `fix lemma` commit(s)).
`test_all_tokens_claimed` still displayed a problem with the `<commands>` cell after processing node 96 but I think we should merge this change anyway  - the new rule is clearer and more focused and it greatly improved performance on some isolated tests that took very long to simplify before.

PR #648, opened 2024-06-12
use optimization level 3 for llvm backend

dwightguth:
We pass -O3 to kompile instead of -O2 in order to decrease time spent doing pattern matching in the llvm backend.


jberthold (approved):
Makes sense (assuming the optimisations are semantics-preserving as they should be)

dwightguth:
Yes, the intention is that this optimization change has no impact on the transition system, and we've pretty much ironed out the rest of the bugs at this point. It makes the binary bigger, but not overly so.

PR #651, opened 2024-06-14
Definedness improvements

bbyalcinkaya:
* Mark unary int ops and int extension ops as `total`.
* Check definedness of `br_table` and add `preserves-definedness`


pykwasm/src/pykwasm/kdist/wasm-semantics/numeric.md
```diff
@@ -115,9 +115,9 @@ module WASM-NUMERIC
 `*UnOp` takes one oprand and returns a `Val`.
 
 ```k
-    syntax Val ::= IValType "." IUnOp   Int   [klabel(intUnOp)    , function]
+    syntax Val ::= IValType "." IUnOp   Int   [klabel(intUnOp)    , function, total]
```
virgil-serbanuta, line 118: Hmmm... Is this really a total function? Let's take the following rule:
```
rule ITYPE . clz    I1 => < ITYPE > #width(ITYPE) -Int #minWidth(I1)
```
which calls `#minWidth(I1)`, which is _NOT_ marked as total and which is defined by the following two rules:
```
    rule #minWidth(0) => 0
    rule #minWidth(N) => 1 +Int #minWidth(N >>Int 1)                                 requires N =/=Int 0
```
which, I'm guessing, will create an infinite loop when `N <Int 0`. If this is correct, then `IValType "." IUnOp Int` is not defined when applied to a type, `clz` and a negative number. Even more, it can't even be a total function for which we didn't write all rules.

So, could you check to see what happens when applied to `clz` and a negative number?

pykwasm/src/pykwasm/kdist/wasm-semantics/wasm.md
```diff
@@ -520,6 +520,11 @@ Note that, unlike in the WebAssembly specification document, we do not need the
  // ---------------------------------------------------------------------
     rule <instrs> #br_table(ES) => #br(#getInts(ES, minInt(VAL, #lenInts(ES) -Int 1))) ... </instrs>
          <valstack> <i32> VAL : VALSTACK => VALSTACK </valstack>
+      requires 0 <=Int VAL
+      [preserves-definedness]
+      // preserves-definedness:
+      // - 0 <= VAL and minInt(VAL, #lenInts(ES) -Int 1) ensures #getInts(_) is defined
```
jberthold, line 526: Is `ES` guaranteed to be non-empty here? Otherwise we'd need a `#lenInts(ES) >Int 0`, too.

pykwasm/src/pykwasm/kdist/wasm-semantics/numeric.md
jberthold: Hmm, that seems more like a problem with `#minWidth` than with the `clz` instruction (which I would think should be total in one way or the other, as it needs to have a result in WASM anyway, right?).
But `N >>Int 1` is still going to reduce `N` to zero eventually, even if `N` is negative, so it might still be OK?

Would it be safe to add a rule
```
   rule ITYPE . clz I1 => 0 requires I1 <Int 0
```
(with the rationale that `I1` in 2-s complement must start with a `1` if it is negative)

pykwasm/src/pykwasm/kdist/wasm-semantics/wasm.md
bbyalcinkaya: It is somehow guaranteed by an implementation choice in our semantics. In the official semantics this instruction has two fields: `#br_table(LIST, DEFAULT)`. If the index is not in bounds, `DEFAULT` value is returned. In our syntax, all the arguments are parsed as a single list so the list is non-empty if the wasm code is valid.

Anyway, since non-emptiness is not ensured by the data structure, it would be better if we add the side condition.

pykwasm/src/pykwasm/kdist/wasm-semantics/numeric.md
```diff
@@ -136,18 +136,18 @@ Note: The actual `ctz` operator considers the integer 0 to have *all* zero-bits,
...
-    syntax Int ::= #minWidth ( Int ) [function]
-                 | #ctz      ( Int ) [function]
-                 | #popcnt   ( Int ) [function]
+    syntax Int ::= #minWidth ( Int ) [function, total]
+                 | #ctz      ( Int ) [function, total]
+                 | #popcnt   ( Int ) [function, total]
  // -------------------------------------------
-    rule #minWidth(0) => 0
-    rule #minWidth(N) => 1 +Int #minWidth(N >>Int 1)                                 requires N =/=Int 0
+    rule #minWidth(N) => 0                                                           requires N <=Int 0
```
jberthold, line 143: That will of course create a problem with `clz` again... e.g., 
`int32.clz -1 => #width(i32) -Int #minWidth(-1) => 4 -Int 0 => 4` which is not correct (should be 0 if `-1` is `11...111`)

pykwasm/src/pykwasm/kdist/wasm-semantics/numeric.md
```diff
@@ -136,18 +136,18 @@ Note: The actual `ctz` operator considers the integer 0 to have *all* zero-bits,
...
+    rule #minWidth(N) => 1 +Int #minWidth(N >>Int 1)                                 requires N >Int 0
 
-    rule #ctz(0) => 0
-    rule #ctz(N) => #if N modInt 2 ==Int 1 #then 0 #else 1 +Int #ctz(N >>Int 1) #fi  requires N =/=Int 0
+    rule #ctz(N) => 0                                                                requires N <=Int 0
+    rule #ctz(N) => #if N modInt 2 ==Int 1 #then 0 #else 1 +Int #ctz(N >>Int 1) #fi  requires N >Int 0
```
jberthold, line 147: AFAICT this function was already total before the change (the right-shift will eventually produce a `1` in the LSb unless the number is indeed `0`).
Should we keep the `modInt` or use `&Int`?
```suggestion
    rule #ctz(N) => 0                      requires N ==Int 0
    rule #ctz(N) => 0                      requires N &Int 1 ==Int 1
    rule #ctz(N) => 1 +Int #ctz(N >>Int 1) [owise]
```

pykwasm/src/pykwasm/kdist/wasm-semantics/numeric.md
```diff
@@ -136,18 +136,18 @@ Note: The actual `ctz` operator considers the integer 0 to have *all* zero-bits,
...
 
-    rule #popcnt(0) => 0
-    rule #popcnt(N) => #bool(N modInt 2 ==Int 1) +Int #popcnt(N >>Int 1)             requires N =/=Int 0
+    rule #popcnt(N) => 0                                                             requires N <=Int 0
+    rule #popcnt(N) => #bool(N modInt 2 ==Int 1) +Int #popcnt(N >>Int 1)             requires N >Int 0
```
jberthold, line 150: If we had size information we could use `<<Int` here, which will indeed eventually produce a straight `0` for both positive and negative `N`... :thinking:

pykwasm/src/pykwasm/kdist/wasm-semantics/numeric.md
```diff
@@ -136,18 +136,18 @@ Note: The actual `ctz` operator considers the integer 0 to have *all* zero-bits,
...
+    rule #minWidth(N) => 1 +Int #minWidth(N >>Int 1)                                 requires N >Int 0
 
-    rule #ctz(0) => 0
-    rule #ctz(N) => #if N modInt 2 ==Int 1 #then 0 #else 1 +Int #ctz(N >>Int 1) #fi  requires N =/=Int 0
+    rule #ctz(N) => 0                                                                requires N <=Int 0
+    rule #ctz(N) => #if N modInt 2 ==Int 1 #then 0 #else 1 +Int #ctz(N >>Int 1) #fi  requires N >Int 0
 
-    rule #popcnt(0) => 0
-    rule #popcnt(N) => #bool(N modInt 2 ==Int 1) +Int #popcnt(N >>Int 1)             requires N =/=Int 0
+    rule #popcnt(N) => 0                                                             requires N <=Int 0
```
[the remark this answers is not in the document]
bbyalcinkaya: @virgil-serbanuta @jberthold In the official wasm spec, these functions are defined only for non-neative numbers (undefined for negative numbers) because integer values are kept as uninterpreted integers. This means that the value in `<i32|i64> N` is always non-negative and converted to a signed int depending on the instruction's signedness. 

* [Values](https://webassembly.github.io/spec/core/exec/runtime.html#values)
* [Integer values](https://webassembly.github.io/spec/core/syntax/values.html#syntax-int)

You can check out the implementation of [`lt_s`](https://github.com/runtimeverification/wasm-semantics/blob/master/pykwasm/src/pykwasm/kdist/wasm-semantics/numeric.md?plain=1#L387) for how unsigned arguments are processed as signed integers using `#signed(_,_)`. 

Ideally I would prefer to use natural numbers instead of `Int` but we don't have built-in natural numbers.

For these reasons, I decided to return a default value of `0` for negative numbers in these functions to keep the implementation as simple as possible. Maybe to be more defensive I can add a check to the instruction's rule to reject negative numbers.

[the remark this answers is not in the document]
jberthold: Maybe the best way forward is indeed to leave these 3 functions the way they were proposed and protect each call site to ensure the argument is non-negative.

One could also make the 3 functions return `-1` for negative arguments, which makes it obvious that the result is invalid (but of course does not protect anything).

jberthold:
FYI I am running a test with these changes (even though the calls are not protected against arguments being negative yet).
My test includes another small change ([marking the `load` instruction for signed result as preserving definedness](https://github.com/runtimeverification/wasm-semantics/commit/8463e8201e37d2e02ed476da4c49e5bf6eb47a13)). Maybe you could consider adding it to this PR, or I can open another one.

pykwasm/src/pykwasm/kdist/wasm-semantics/numeric.md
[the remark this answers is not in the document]
bbyalcinkaya: List of updates after your reviews:
* `#minWidth`, `#ctz`, `#popcnt` returns `-1` for negative numbers to indicate a failure. I think this will be better than `0` when debugging if anything goes wrong.
* The `T:IValType . OP:IUnOp I1` function returns `undefined` for negative numbers.
* Replaced `modInt 2` with `&Int 1` for `&Int` is a total function.

virgil-serbanuta (approved):

PR #664, opened 2024-06-28
Clean up nix code

Baltoli:
Now that Pyk and K have aligned their respective versions of Nixpkgs, we can begin to clean up some of the Nix code that was previously required to work around them being separated.


goodlyrottenapple (approved):
LGTM. One thing we could also tighten up would be to make everything depend on the same rv_utils down the chain. Even though we point to the same rv_utils independently, in the flake lock, we have 4 copies. Having only one copy should prevent subtle bugs where once copy could potentially be different to the rest...

PR #670, opened 2024-07-08
Support `memory.fill`

gtrepta:
This adds the semantics for the `memory.fill` instruction.

In the spec they desugar the instruction into a sequence of `i32.store8` instructions. For efficiency, we just use the `#setBytesRange` function.

https://webassembly.github.io/spec/core/exec/instructions.html#xref-syntax-instructions-syntax-instr-memory-mathsf-memory-fill


pykwasm/src/pykwasm/kdist/wasm-semantics/wasm.md
```diff
@@ -1568,6 +1569,51 @@ The maximum of table size is 2^32 bytes.
     rule #maxTableSize()  => 4294967296
 ```
 
+`fill` fills a contiguous section of memory with a value.
+When the section specified goes beyond the bounds of the memory region, that causes a trap.
+If the section has length 0, nothing happens.
+The spec states that this is really a sequence of `i32.store8` instructions, but we use `#setBytesRange` here.
+
```
virgil-serbanuta, line 1576: Would it be worth adding a link to the spec here?

virgil-serbanuta (approved):
LGTM.

There is a difference from the spec in that the memory state may be slightly different when the trap executes, but I guess that's reasonable. It may be worth adding a note about that.

pykwasm/src/pykwasm/kdist/wasm-semantics/wasm.md
gtrepta: I can add links to each of the memory instructions in this sections, could be helpful.

>There is a difference from the spec in that the memory state may be slightly different when the trap executes, but I guess that's reasonable. It may be worth adding a note about that.

I'm not sure I understand what you're talking about here. Do you think some of the `i32.store8` could complete successfully before a trap? I don't think that's the case, because the validation on the initial values of `d` and `n` should guarantee the validation for all of the `store8` instructions.

virgil-serbanuta: > > There is a difference from the spec in that the memory state may be slightly different when the trap executes, but I guess that's reasonable. It may be worth adding a note about that.
> 
> I'm not sure I understand what you're talking about here. Do you think some of the `i32.store8` could complete successfully before a trap? I don't think that's the case, because the validation on the initial values of `d` and `n` should guarantee the validation for all of the `store8` instructions.

I misread the spec, you're right.

PR #680, opened 2024-07-17
Fix table imports and checks for data segments

gtrepta:
When instantiating a data segment, we weren't checking that the segment doesn't go out of bounds of the memory in the module.

When importing a table, we weren't using the `<nextTabIdx>` cell to allow multiple imports.

This change fixes these two issues.


virgil-serbanuta (approved):
FWIW, tests would be nice.

gtrepta:
> FWIW, tests would be nice.

The `linking.wast` test revealed these two issues and I'm removing it from the unsupported-llvm list in this PR. Introducing some more unit-like tests with pytest would be good, but I probably can't start that effort now since I'll be moving to soroban soon.

PR #696, opened 2024-09-04
Minor nix cleanup

goodlyrottenapple:
I noticed we were using the wrong version of poetry2nix in this flake. Confusingly, the default nixpkgs repository contains a version of the poetry2nix library. However, we want to use the poetry2nix library that k-framework uses, which ultimately points to the community version of the library: https://github.com/nix-community/poetry2nix


tothtamas28 (approved):
LGTM, please add a short description to the PR.

tothtamas28 (approved):

PR #714, opened 2024-12-13
Fix prove tests

traiansf:
- [ ] Fix (prove) tests (broken by refactoring `<locals>` to hold a `List` instead of a `Map`) so that `make test` passes
- [ ] Add Makefile targets to (only) build specific targets


traiansf:
Sorry, opened in the wrong repo

PR #717, opened 2025-01-22
Add `preserves-definedness` attributes

bbyalcinkaya:


ehildenb (approved):

ehildenb:
I checked,  and it looks ilke in EVM semantics we also use `preservse-definedness` attribute for any rules modifying cells in collections, so I guess it makes sense here too!

